PR #26, opened 2022-06-08
Coinbase transactions should not be spendable until after 50 blocks.

reddink:
Added some fixes to the recent updates of coinbase maturity


reddink:
@barrystyle 
just a couple of fixes for the coinbase maturity changes.
I was also thinking to parameterise the additional 20 blocks for the spendability of coinbase|coinstake txs

barrystyle:
Corrected the issue in PR#27.

PR #35, opened 2022-11-16
qt: Show all addresses without label in a less visible color

lateminer:
Staking address should also be in gray color if no label is assigned.


reddink:
ACK

That looks better

PR #36, opened 2022-11-18
[CI] enable CI for reddcoin

reddink:
Enable CI


reddink:
First pass enabling CI,
Still a few fix-ups to do in code

PR #55, opened 2023-01-15
Check if a wallet is loaded before working with stake thread

lateminer:
It is needed to check if there is actually a wallet available, otherwise there will be segfault.


reddink:
Good catch,
Will merge this in shortly

PR #56, opened 2023-01-16
Avoid getting into infinite loops in staking thread

lateminer:
If the wallet is locked, the thread is stuck in the loop, so it is needed to check for ShutdownRequested() and stop the thread after the signal received.


reddink:
Nice,
Actually saw this issue myself yesterday on testnet
Will merge this in shortly

PR #57, opened 2023-02-01
Do not start staking thread if private keys are disabled

lateminer:
It is pointless to start staking thread if a wallet doesn't contain private keys.


reddink:
Fair call,
Merging this in

PR #320, opened 2025-06-12
Update bitcoinunits.cpp

meghasurya07:
I suggest the name Reddoshi (rsh) if you would like it review the changes or lets keep the original


reddink:
I prefer to keep the original sat = satoshi rather than changing to something new.
The main reason being, that almost all people are familiar with that term (satoshi) and being the smallest unit

meghasurya07:
@reddink sure no worries
i wonder why no one is worrying about the CI tests

reddink:
I am aware of the ci failures.
Just haven't gotten back to the next batch i want to work on.

PR #437, opened 2026-07-28
ci: fetch Reddcoin's previous release, not Bitcoin's

reddink:
# ci: fetch Reddcoin's previous release, not Bitcoin's

## Summary

The `previous releases, qt5, DEBUG` job fails:

```
FileNotFoundError: [Errno 2] No such file or directory:
'/home/runner/work/reddcoin/reddcoin/releases/x86_64-pc-linux-gnu/v4.22.9/bin/reddcoind'
```

## Cause

Nothing had ever been asked to produce that file. The previous-release machinery is still entirely Bitcoin's:

```
export PREVIOUS_RELEASES_TO_DOWNLOAD="v0.15.2 v0.16.3 v0.17.2 v0.18.1 v0.19.1 v0.20.1"
```

and the job duly fetches them:

```
Releases directory: /home/runner/work/reddcoin/reddcoin/releases/x86_64-pc-linux-gnu
Fetching: https://bitcoincore.org/bin/bitcoin-core-0.15.2/bitcoin-0.15.2-x86_64-linux-gnu.tar.gz
```

Meanwhile `TEST_RUNNER_EXTRA` forces `--previous-releases` on, and the compat tests want v4.22.9. So the job downloads six Bitcoin releases nothing will use, and none of the one thing it needs.

`get_previous_releases.py` was half converted. Its build path already moves `reddcoind`, `reddcoin-cli` and `reddcoin-tx` into place at the end, but still clones `bitcoin/bitcoin` to get them. Its download path is bitcoincore.org and Bitcoin's checksums throughout.

## Changes

The shape stays upstream's, since that is worth keeping: fetch an archive for the tag, check it against a recorded sum, extract it into place.

**Download from our own release assets.** The archive is `reddcoin-<version>-<platform>.tar.gz`, taken from the release for the tag. `PREVIOUS_RELEASES_URL` overrides the base for a mirror or a staging location. `curl` follows redirects now, which release assets need.

**Clone Reddcoin in the build path**, so the fallback matches what it then packages.

**Empty the checksum table** of Bitcoin's sums, to be filled as archives are uploaded.

**Ask for v4.22.9.3** in the qt5 environment, for the reason below.

## Which release, and why it is not v4.22.9

The build the compat tests need is not the released v4.22.9, which cannot run regtest at all. It is that release plus the five commits on the `v4.22.9-regtest` branch, which enable the CSV, SegWit and CLTV deployments there.

Giving that its own version is cheaper than it looks. The framework already builds the release directory from four components and only strips a trailing zero:

```python
're.sub(r'\.0$', '', 'v{}.{}.{}.{}'.format(...))'
```

so `4220900` was always v4.22.9 build 0, and `4220903` is the same release at build 3, resolving to `releases/v4.22.9.3/bin`. Builds 1 and 2 are already spoken for by earlier candidates. The fourth component is already part of the scheme rather than something added for this, and nothing in the framework changes. `configure.ac` is left alone, so the binary still reports v4.22.9; nothing reads that, the directory name is what the framework keys on.

The prose in those tests still says v4.22.9 where it describes behaviour, which stays accurate. The build differs from the release only in what it will start, not in how it validates.

## The archive this needs

Nothing needs packaging for it. Guix already writes exactly what the fetch expects:

```
guix-build-<version>/output/<host>/reddcoin-<version>-<host>.tar.gz
```

with a `reddcoin-<version>` directory inside it, since `DISTNAME` is `reddcoin-${VERSION}` and the binaries are staged under `installed/${DISTNAME}`. Pass `FORCE_VERSION=4.22.9.3` so it is named for the release rather than for the commit built, `HOSTS=x86_64-linux-gnu` to build only the one host the job needs, and upload the result.

Guix also settles the harder of the two requirements. This job builds in ubuntu:18.04, which carries glibc 2.27, and a tree built on a current host will not start there: the binaries to hand want GLIBC_2.34. Guix builds Linux hosts against glibc 2.24, comfortably below what the container provides.

The other requirement is one only the source tree can meet: it has to be a build that can run regtest at all, which the released v4.22.9 cannot. That is what the five commits on the `v4.22.9-regtest` branch are for, enabling CSV, SegWit and CLTV signaling there. Guix should be run against that tree.

## Testing

The whole chain is verified end to end, against a local server standing in for the release host and an archive built from the existing `v4.22.9-regtest` tree. It fetches, matches the checksum, and extracts to `releases/<target>/v4.22.9.3/bin/` with `reddcoind`, `reddcoin-cli` and `reddcoin-tx` in it.

`feature_compat_p2p_sync` then passes against it, which is the part worth having: it resolves `4220903` to that directory and starts the old binary. A missing one raises, which is how this failed in CI, so passing means it ran.

That archive was a host build, so it stands in for the shape and not for the contents; a Guix one differs in what it links against, not in how it is named or laid out.

The job itself cannot pass until an archive is uploaded, and the checksum table is deliberately empty until one is. Fetching an asset that is not there fails on the 404 check, and fetching one whose sum is unrecorded fails on the checksum, so neither can pass silently.

### DIFF
--- a/ci/test/00_setup_env_native_qt5.sh
+++ b/ci/test/00_setup_env_native_qt5.sh
@@ -14,6 +14,6 @@ export TEST_RUNNER_EXTRA="--previous-releases --coverage --extended --exclude fe
 export RUN_UNIT_TESTS_SEQUENTIAL="true"
 export RUN_UNIT_TESTS="false"
 export GOAL="install"
-export PREVIOUS_RELEASES_TO_DOWNLOAD="v0.15.2 v0.16.3 v0.17.2 v0.18.1 v0.19.1 v0.20.1"
+export PREVIOUS_RELEASES_TO_DOWNLOAD="v4.22.9"  # the release the compat tests run against
 export BITCOIN_CONFIG="--enable-zmq --with-libs=no --with-gui=qt5 --enable-glibc-back-compat --enable-reduce-exports
 --enable-debug --disable-fuzz-binary  CFLAGS=\"-g0 -O2 -funsigned-char\" CXXFLAGS=\"-g0 -O2 -funsigned-char\""
--- a/ci/test/00_setup_env_native_qt5.sh
+++ b/ci/test/00_setup_env_native_qt5.sh
@@ -14,6 +14,6 @@ export TEST_RUNNER_EXTRA="--previous-releases --coverage --extended --exclude fe
 export RUN_UNIT_TESTS_SEQUENTIAL="true"
 export RUN_UNIT_TESTS="false"
 export GOAL="install"
-export PREVIOUS_RELEASES_TO_DOWNLOAD="v4.22.9"  # the release the compat tests run against
+export PREVIOUS_RELEASES_TO_DOWNLOAD="v4.22.9.3"  # the release the compat tests run against
 export BITCOIN_CONFIG="--enable-zmq --with-libs=no --with-gui=qt5 --enable-glibc-back-compat --enable-reduce-exports
 --enable-debug --disable-fuzz-binary  CFLAGS=\"-g0 -O2 -funsigned-char\" CXXFLAGS=\"-g0 -O2 -funsigned-char\""
--- a/test/functional/feature_compat_p2p_sync.py
+++ b/test/functional/feature_compat_p2p_sync.py
@@ -26,7 +26,7 @@ the node falls back to a PoW template which is rejected past nLastPowHeight).
     Disconnect, node1 mines 50 PoS blocks, reconnect, node0 catches up.
     Repeat once more to test repeated long-gap sync.
 
-Requires previous releases (v4.22.9 binary), see test/README.md.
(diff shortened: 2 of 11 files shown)


reddink:
Closing in favour of folding this into the RED-43 CI migration branch, where it belongs alongside the rest of the CI work rather than landing ahead of it.

Both commits are already on `red-43-circleci-migration`, patch-identical to the ones here (verified with `git patch-id --stable`):

| this PR | on red-43 |
| --- | --- |
| `cac9dd29b4` ci: fetch Reddcoin's previous release, not Bitcoin's | `bf5c5d3c01` |
| `d938ae6647` test: point the compat tests at the v4.22.9.3 previous release | `c04eb68cdd` |

So no content is lost by closing, and nothing needs cherry-picking. The previous-release fetch will ship as part of the RED-43 pipeline PR.

Leaving the `ci/previous-release-fetch` branch in place until RED-43 lands, as a fallback.

PR #477, opened 2026-08-07
ci: remove the Cirrus configuration now GitHub Actions has replaced it

reddink:
Closes out the last config item on [RED-43](https://reddink.youtrack.cloud/issue/RED-43).

Cirrus CI stopped running on 2026-06-01 after the company was acquired, so nothing has triggered from `.cirrus.yml` since. The GitHub Actions pipeline that replaced it landed in #455 and has been the only thing running CI ever since, with all fourteen Cirrus tasks ported and their job names kept identical to the tasks they replace.

The file was deliberately kept at the time so the two could be compared while the port settled. That comparison is done.

## Nothing depends on it

Checked before removing:

- No script under `ci/` or `test/lint/` reads it. The build logic was always provider-agnostic, selected by `FILE_ENV`, which is why the migration was a config translation rather than a script rewrite.
- The remaining `cirrus` mentions in `.github/` are comments recording where each job came from, for example *"Ported from .cirrus.yml after Cirrus CI stopped running on 2026-06-01"* and the notes on `cirruslabs/cache` in the cache actions. Those are useful provenance and are kept.
- `src/secp256k1/.cirrus.yml` is the subtree's own CI and is untouched.
- Mentions in `doc/release-notes/` are historical record and are left alone.

## Also in this change

`.editorconfig` had a comment naming `.cirrus.yml` as its example `*.yml` file. It is repointed at `.github/workflows/ci.yml`, and drops `.fuzzbuzz.yml`, which this tree does not have. The rule itself is unchanged and still applies to the workflow files. Without this the comment would name a file that no longer exists.

## `.circleci/` needs no action

RED-43's checklist also lists deciding the fate of the CircleCI artifacts. There is nothing to remove: `.circleci/` was never merged to `develop`. That work lived only on the shelved `red-43-circleci-migration` branch, which was abandoned when running it tripped CircleCI's Acceptable Use Policy clause on *"mining or demonstrating proof-of-work or other proof by use of resources for any cryptocurrency or blockchain"*. The branch is merged, so `34d44e9bb6` remains reachable in history if the structural reference is ever wanted.

## Testing

Nothing to run. The removed file is inert config for a service that no longer executes, and the only code change is a comment. The pipeline that matters is exercised by this PR itself.


reddink:
Withdrawing this for now. Straightforward to re-cut if we come back to it: the change is just the deletion of `.cirrus.yml` plus a one-line `.editorconfig` comment repoint.

PR #503, opened 2026-09-01
node: verify TLS certificates and send SNI in the update check

reddink:
Fixes the update check accepting any certificate from any party able to intercept the connection.

## Problem

`HttpsFetcher` constructed an ssl context and never configured it, so there were no trust anchors, verification was never requested, and the hostname was never checked against the certificate. The connection was encrypted but unauthenticated.

The most useful thing that buys an attacker is **silently suppressing upgrade notices**, which is the wrong failure mode for the mechanism whose job is to get security fixes onto user machines. It also let them inject an arbitrary version string into the comparison. The blast radius is limited today because `officialDownloadLink` is assembled from compiled-in constants, but that limit disappears as soon as the client downloads anything, which is why this gates RED-98.

## A correction to the issue and the design doc

Both say Linux is fine and only Windows and macOS need care. That is wrong, and it makes the "obvious" fix actively harmful.

depends builds OpenSSL with `--openssldir=$(host_prefix)/etc/openssl`, which bakes the **builder's absolute path** into libcrypto. Read back out of the built `libcrypto.a`:

```
OPENSSLDIR: "/home/john/.../reddcoin/depends/x86_64-pc-linux-gnu/etc/openssl"
```

That directory exists on no user's machine on **any** platform. So a fix that just calls `set_default_verify_paths()` plus `verify_peer` would find zero anchors and break the update check on every release binary, turning a silent weakness into a visible outage everywhere.

Two smaller corrections:

- `host_name_verification`, which the issue and the design doc both prescribe, **arrived in Boost 1.73**. depends pins 1.71, which has only `rfc2818_verification`. The prescribed call would not compile.
- OpenSSL's `winstore` was explicitly disabled in depends (`no-winstore`), so it had to be enabled here.

## Fix

Anchors come from the operating system, in new `src/node/ca_store.{h,cpp}`:

| Platform | Source |
| --- | --- |
| Windows | OpenSSL's winstore loader over the system ROOT store |
| macOS | system trust store via Security.framework |
| Everything else | `SSL_CERT_FILE` / `SSL_CERT_DIR`, else the first well-known distribution bundle that exists |

Taking anchors from the platform means the trust list follows the OS's own updates and revocations, and leaves no CA list in this repository for somebody to remember to refresh. That is the same reasoning that led the epic to reject pinning a release key.

The fetcher then sets `verify_peer`, an `rfc2818_verification` callback bound to the host, and SNI via `SSL_set_tlsext_host_name`.

**Failing to obtain anchors is fatal to the connection**, not a downgrade to an unverified fetch. It surfaces through the existing `errors` field.

One subtlety worth reviewing: the loader checks the compiled-in location *exists* before falling back to it, because `SSL_CTX_set_default_verify_paths` only registers the lookups and reports success without touching the filesystem. Trusting its return value would report an empty trust store as a working one, which is precisely the depends case this is meant to catch.

## Build changes

- `depends/packages/openssl.mk`: drop `no-winstore`.
- `configure.ac`: link `crypt32` on Windows, `Security` and `CoreFoundation` on macOS. Both frameworks are present in the depends SDK.

## Testing

Verified against live endpoints, using a harness around the real `ca_store.cpp` and the exact context configuration the fetcher applies:

```
PASS  accept api.github.com          -> handshake succeeded
PASS  accept download.reddcoin.com   -> handshake succeeded
PASS  reject expired.badssl.com          -> certificate verify failed
PASS  reject wrong.host.badssl.com       -> certificate verify failed
PASS  reject self-signed.badssl.com      -> certificate verify failed
PASS  reject untrusted-root.badssl.com   -> certificate verify failed
```

Fail-closed confirmed separately: with `SSL_CERT_FILE` pointing at a nonexistent path the fetch refuses to connect at all rather than proceeding unverified, and with it pointing at a real bundle it succeeds.

`src/node/ca_store.cpp` and `src/node/update_check.cpp` compile clean.

**Not verified:** the Windows and macOS paths are not compile-tested here, as this has no cross toolchain built. The winstore call and the Security.framework enumeration need a check on a real cross build before this is called done on all hosts.

YouTrack: https://reddink.youtrack.cloud/issue/RED-99
Phase 0 of https://reddink.youtrack.cloud/issue/RED-98

Backport to the 4.22.9 release line in a companion PR, which also carries the OpenSSL 1.1.1s to 3.5.7 bump that branch needs first.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -652,6 +652,10 @@ case $host in
      AC_CHECK_LIB([ws2_32],   [WSAStartup],, AC_MSG_ERROR(libws2_32 missing))
      AC_CHECK_LIB([shlwapi],  [PathRemoveFileSpecW],, AC_MSG_ERROR(libshlwapi missing))
      AC_CHECK_LIB([iphlpapi], [GetAdaptersAddresses],, AC_MSG_ERROR(libiphlpapi missing))
+     dnl OpenSSL's winstore loader reads the system ROOT certificate store, which
+     dnl is how the update check gets its TLS trust anchors on Windows. depends
+     dnl must not be configured with no-winstore.
+     AC_CHECK_LIB([crypt32],  [CertOpenStore],, AC_MSG_ERROR(libcrypt32 missing))
 
      dnl -static is interpreted by libtool, where it has a different meaning.
      dnl In libtool-speak, it's -all-static.
@@ -734,6 +738,9 @@ case $host in
 
      AX_CHECK_LINK_FLAG([[-Wl,-headerpad_max_install_names]], [LDFLAGS="$LDFLAGS -Wl,-headerpad_max_install_names"],, [[$LDFLAG_WERROR]])
      CPPFLAGS="$CPPFLAGS -DMAC_OSX -DOBJC_OLD_DISPATCH_PROTOTYPES=0"
+     dnl The update check reads its TLS trust anchors from the system trust
+     dnl store, which OpenSSL has no loader for on this platform.
+     LIBS="$LIBS -framework Security -framework CoreFoundation"
      OBJCXXFLAGS="$CXXFLAGS"
      ;;
    *android*)
--- a/contrib/devtools/symbol-check.py
+++ b/contrib/devtools/symbol-check.py
@@ -112,6 +112,7 @@ MACHO_ALLOWED_LIBRARIES = {
 
 PE_ALLOWED_LIBRARIES = {
 'ADVAPI32.dll', # security & registry
+'CRYPT32.dll', # certificate store, for the update check's TLS trust anchors
 'IPHLPAPI.DLL', # IP helper API
 'KERNEL32.dll', # win32 base APIs
 'msvcrt.dll', # C standard library for MSVC
--- a/depends/packages/openssl.mk
+++ b/depends/packages/openssl.mk
@@ -37,11 +37,11 @@ $(package)_config_opts+=no-tests
 $(package)_config_opts+=no-unit-test
 $(package)_config_opts+=no-weak-ssl-ciphers
 $(package)_config_opts+=no-whirlpool
-# The 3.x series added a store backed by the Windows certificate store, which
-# calls into crypt32 and so does not link unless that library is added to every
-# Windows target. Nothing here uses it, and capieng above is already disabled
-# for the same reason.
-$(package)_config_opts+=no-winstore
+# The 3.x series added a store backed by the Windows certificate store. It is
+# how the update check gets its TLS trust anchors on Windows, since depends
+# bakes the builder's own path into OPENSSLDIR and so the default verify paths
+# find nothing on the machine that runs the binary. It calls into crypt32,
+# which configure.ac links for every Windows target.
 $(package)_config_opts+=no-zlib
 $(package)_config_opts+=no-zlib-dynamic
 $(package)_config_opts+=$($(package)_cflags) $($(package)_cppflags)
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -187,6 +187,7 @@ BITCOIN_CORE_H = \
   node/blockstorage.h \
   node/coin.h \
   node/coinstats.h \
+  node/ca_store.h \
   node/context.h \
   node/psbt.h \
   node/transaction.h \
@@ -365,6 +366,7 @@ libbitcoin_server_a_SOURCES = \
   node/blockstorage.cpp \
   node/coin.cpp \
   node/coinstats.cpp \
+  node/ca_store.cpp \
   node/context.cpp \
   node/interfaces.cpp \
   node/psbt.cpp \
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -147,6 +147,7 @@ BITCOIN_TESTS =\
   test/txrequest_tests.cpp \
   test/txvalidation_tests.cpp \
   test/uint256_tests.cpp \
+  test/update_check_tests.cpp \
   test/uint512_tests.cpp \
   test/util_tests.cpp \
   test/validation_block_tests.cpp \
--- a/src/node/ca_store.cpp
+++ b/src/node/ca_store.cpp
@@ -1,0 +1,177 @@
+// Copyright (c) 2014-2026 The Reddcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <node/ca_store.h>
+
+#include <openssl/err.h>
+#include <openssl/opensslv.h>
+#include <openssl/ssl.h>
+#include <openssl/x509.h>
+
+#include <cstdlib>
+#include <string>
+
+#if defined(WIN32)
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+#error "Windows builds need OpenSSL 3.0 or later for the winstore certificate loader; build against depends."
+#endif
+#elif defined(MAC_OSX)
+#include <CoreFoundation/CoreFoundation.h>
+#include <Security/Security.h>
+#else
+#include <sys/stat.h>
+#endif
+
+namespace {
+#if defined(MAC_OSX)
+//! Add one DER-encoded certificate to a store.
+//!
+//! A certificate the store already holds is not a failure. Platform trust
+//! stores can list the same root more than once, and OpenSSL reports the
+//! duplicate as an error that would otherwise abort the whole load.
+bool AddDerCertificate(X509_STORE* store, const unsigned char* der, long der_len)
+{
+    // d2i_X509 advances the pointer it is given, so hand it a copy.
+    const unsigned char* pos{der};
+    X509* cert{d2i_X509(nullptr, &pos, der_len)};
+    if (cert == nullptr) {
+        ERR_clear_error();
+        return false;
+    }
+
+    bool ok{true};
+    if (X509_STORE_add_cert(store, cert) != 1) {
+        ok = ERR_GET_REASON(ERR_peek_last_error()) == X509_R_CERT_ALREADY_IN_HASH_TABLE;
+        ERR_clear_error();
+    }
+
+    // X509_STORE_add_cert takes its own reference, so drop ours either way.
+    X509_free(cert);
+    return ok;
+}
+#endif // MAC_OSX
+
+#if !defined(WIN32) && !defined(MAC_OSX)
+//! Bundle locations used by the distributions this is likely to run on. The
+//! first one that exists wins; they hold the same set of public roots.
+const char* const CA_FILE_CANDIDATES[]{
+    "/etc/ssl/certs/ca-certificates.crt",                // Debian, Ubuntu, Gentoo, Alpine
+    "/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora, RHEL, CentOS
+    "/etc/ssl/ca-bundle.pem",                            // openSUSE
+    "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS 7 and later
+    "/etc/ssl/cert.pem",                                 // FreeBSD, OpenBSD, Alpine
+};
+
+//! Hashed certificate directories, used only if no bundle file is found.
+const char* const CA_DIR_CANDIDATES[]{
+    "/etc/ssl/certs",
+    "/etc/pki/tls/certs",
+};
+
+bool PathExists(const char* path)
+{
+    struct stat sb;
+    return ::stat(path, &sb) == 0;
+}
+#endif // !WIN32 && !MAC_OSX
+} // namespace
+
+std::string node::LoadTrustedCACertificates(SSL_CTX* ssl_ctx)
+{
+    if (ssl_ctx == nullptr) return "No TLS context to load certificates into";
+
+#if defined(WIN32)
+    // OpenSSL 3.2 and later expose the Windows ROOT store through the OSSL_STORE
+    // URI below. It reads the same anchors the rest of the system trusts, so it
+    // picks up enterprise roots and administrator revocations without any
+    // enumeration code here. depends must not be configured with no-winstore.
+    if (SSL_CTX_load_verify_store(ssl_ctx, "org.openssl.winstore://") != 1) {
+        ERR_clear_error();
+        return "Could not read the Windows system certificate store";
+    }
+    return "";
+
+#elif defined(MAC_OSX)
+    // OpenSSL has no equivalent loader for the macOS keychain, so the anchors
+    // are copied across one at a time.
+    CFArrayRef anchors{nullptr};
+    if (SecTrustCopyAnchorCertificates(&anchors) != errSecSuccess || anchors == nullptr) {
+        return "Could not read the macOS system trust store";
+    }
+
+    X509_STORE* store{SSL_CTX_get_cert_store(ssl_ctx)};
+    if (store == nullptr) {
+        CFRelease(anchors);
+        return "No TLS certificate store to load anchors into";
+    }
+
+    long loaded{0};
+    const CFIndex count{CFArrayGetCount(anchors)};
+    for (CFIndex i = 0; i < count; ++i) {
+        const void* elem{CFArrayGetValueAtIndex(anchors, i)};
+        if (elem == nullptr) continue;
+        SecCertificateRef cert{reinterpret_cast<SecCertificateRef>(const_cast<void*>(elem))};
+
+        CFDataRef der{SecCertificateCopyData(cert)};
+        if (der == nullptr) continue;
+        if (AddDerCertificate(store, CFDataGetBytePtr(der), static_cast<long>(CFDataGetLength(der)))) {
+            ++loaded;
+        }
+        CFRelease(der);
+    }
+    CFRelease(anchors);
+
+    if (loaded == 0) return "The macOS system trust store contained no usable certificates";
+    return "";
+
+#else
+    // An explicit environment setting is honoured first, so a container or a
+    // distribution that keeps its bundle somewhere unusual can point at it
+    // without a rebuild. These are the same two variables OpenSSL's own default
+    // verify paths consult.
+    const char* const env_file{std::getenv("SSL_CERT_FILE")};
+    const char* const env_dir{std::getenv("SSL_CERT_DIR")};
+    if ((env_file != nullptr && *env_file != '\0') || (env_dir != nullptr && *env_dir != '\0')) {
+        if (SSL_CTX_load_verify_locations(ssl_ctx,
+                                          (env_file != nullptr && *env_file != '\0') ? env_file : nullptr,
+                                          (env_dir != nullptr && *env_dir != '\0') ? env_dir : nullptr) == 1) {
+            return "";
+        }
+        ERR_clear_error();
+        return "SSL_CERT_FILE or SSL_CERT_DIR is set but no certificates could be read from it";
+    }
+
+    for (const char* const candidate : CA_FILE_CANDIDATES) {
+        if (!PathExists(candidate)) continue;
+        if (SSL_CTX_load_verify_locations(ssl_ctx, candidate, nullptr) == 1) return "";
+        ERR_clear_error();
+    }
+
+    for (const char* const candidate : CA_DIR_CANDIDATES) {
+        if (!PathExists(candidate)) continue;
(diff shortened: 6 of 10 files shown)


reddink:
**Runtime verified end-to-end in a real `reddcoin-qt` build.**

Self-compiled Linux `reddcoin-qt` at `ddeacf898a`, statically linked against the depends OpenSSL, running on mainnet:

```
$ reddcoin-cli checkupdates
{
  "localversion": "4.22.9.4",
  "remoteversion": "4.22.9.4",
  "updateavailable": false,
  "message": "You're running the most recent version of Reddcoin Core (4.22.9.4)",
  "warning": "",
  "officialDownloadLink": "https://download.reddcoin.com/bin/reddcoin-core-4.22.9.4",
  "errors": ""
}
```

This is stronger evidence than it looks, for two reasons.

**1. Success proves verification was on, not bypassed.** `HttpsFetcher`'s constructor is linear and has no branch that skips verification: load anchors (throws on failure) -> `set_verify_mode(verify_peer)` -> `set_verify_callback` -> SNI -> handshake. There is no setting that disables it. So a populated `remoteversion` with an empty `errors` means the anchor load succeeded, `verify_peer` was active, and `api.github.com` chained to a trusted root and matched the hostname. SNI is likewise exercised, since api.github.com sits behind a CDN.

**2. It exercises the exact case the naive fix would have broken.** The binary links the depends OpenSSL statically, so `OPENSSLDIR` is the builder's own path and `set_default_verify_paths()` finds nothing. The fetch succeeded only because `ca_store.cpp` fell through to the distribution bundle probe. Had this been implemented as the issue described, this build would have reported a CA failure instead of a version.

Combined with the harness results already in the description (expired, wrong-host, self-signed and untrusted-root all rejected; unreadable trust store fails closed), the Linux path is now covered at both ends: it accepts what it should and rejects what it should, in the integrated binary.

**Still not covered:** the Windows and macOS paths remain uncompiled here. A guix build with `HOSTS="x86_64-w64-mingw32 x86_64-apple-darwin18"` is the right gate before merge, and it also exercises the new `crypt32` and Security/CoreFoundation link flags in `configure.ac`.

reddink:
**Windows cross build: compiles and links. Fixed the one gate it hit.**

The guix `x86_64-w64-mingw32` build got all the way through compiling and linking both binaries, so the winstore loader and the new `crypt32` linkage are good on Windows. It then failed the symbol check:

```
Checking Windows dynamic libraries...
CRYPT32.dll is not in ALLOWED_LIBRARIES!
reddcoind.exe: failed DYNAMIC_LIBRARIES
CRYPT32.dll is not in ALLOWED_LIBRARIES!
qt/reddcoin-qt.exe: failed DYNAMIC_LIBRARIES
make: *** [Makefile:18931: check-symbols] Error 1
```

That is a policy gate, not a build failure: linking crypt32 adds `CRYPT32.dll` to the binaries' imports, and `contrib/devtools/symbol-check.py` rejects any Windows DLL not on its allow list. Added it in `a212fc8425`. CRYPT32.dll ships with every supported version of Windows, so it introduces no new runtime requirement.

No matching change is needed for macOS: `MACHO_ALLOWED_LIBRARIES` already permits `Security` and `CoreFoundation`.

This closes most of the "not compile-tested on Windows" caveat in the description. What the cross build now demonstrates:

- `ca_store.cpp` compiles for mingw, so the `SSL_CTX_load_verify_store(..., "org.openssl.winstore://")` path and the `OPENSSL_VERSION_NUMBER` guard are sound.
- Dropping `no-winstore` in depends produces a libcrypto that actually resolves the winstore symbols.
- The `AC_CHECK_LIB([crypt32], [CertOpenStore])` addition to configure.ac works, and crypt32 genuinely ends up linked, which the symbol check itself proves.

**Still outstanding:** the Windows and macOS paths are compiled but not *executed*. Nothing here shows the winstore loader returns real anchors at runtime on Windows, or that the Security.framework enumeration works on macOS. That wants someone running the resulting binaries and calling `checkupdates`, expecting a populated `remoteversion` and an empty `errors`. The macOS cross build has not been run at all yet.

reddink:
**macOS runtime verified.** This closes the last of the "compiled but never executed" caveat for darwin.

Guix cross build at `a212fc842505`, `reddcoin-<ver>-osx64.tar.gz` run on a macOS 10.15.7 (x86_64) VM.

The frameworks are genuinely linked, so the `configure.ac` change works on a real cross build:

```
$ otool -L bin/reddcoind | grep -E "Security|CoreFoundation"
  /System/Library/Frameworks/Security.framework/Versions/A/Security
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
```

`otool -L` shows no dynamic libssl or libcrypto, so OpenSSL is statically linked from depends. That means `OPENSSLDIR` is the builder's path and the Security.framework loader is the *only* possible source of trust anchors on this binary. Nothing else could have supplied them.

```
$ reddcoin-cli -regtest checkupdates
"localversion":"4.22.9.0", "remoteversion":"4.22.9.4",
"updateavailable":true, "errors":""
```

Populated `remoteversion` with empty `errors` means `SecTrustCopyAnchorCertificates` returned usable anchors, `verify_peer` was active, and `api.github.com` chained to a system root and matched the hostname. The version comparison is correct too: the branch reports 4.22.9.0 against a 4.22.9.4 release, so `updateavailable` is true.

Windows also cleared `check-symbols` in the same build after `a212fc8425`, producing its full artifact set and `SHA256SUMS.part`. **Windows is still not runtime-tested** - no Windows VM here - so the winstore loader remains compiled and linked but unexecuted.

---

**One anomaly worth recording, not a TLS problem.**

The very first `checkupdates` after the first daemon start returned every field empty *including* `errors`:

```
{"localversion":"","remoteversion":"","updateavailable":false, ... "errors":""}
```

An empty `errors` rules out the TLS path: every failure in `HttpsFetcher` (CA load, handshake, non-200 status, malformed status line, missing header terminator, timeout) throws, and `CheckForUpdates` turns that into `errors`. So the handshake succeeded and a well-formed HTTP 200 with headers came back; only `obj_response.read()` failed to find `tag_name`, leaving every field at its initial value.

The most likely cause is the read loop silently accepting a short body: `async_read` with `transfer_all()` treats both `eof` and `stream_truncated` as success, so a connection that ends mid-body yields a truncated response that parses to nothing and reports no error. That is the buffering/robustness gap already recorded as blocker B3 in `doc/design/assisted-upgrade.md`, in code this PR does not touch.

It did not reproduce: 7 subsequent calls across two daemon starts, including 4 issued back to back immediately after startup, all returned 4.22.9.4 with empty `errors`. Recording it rather than dismissing it, since a silently truncated body reported as success is worth fixing on its own merits when B3 is addressed.

`curl` from the same host confirmed GitHub serves this endpoint with `Content-Length` and `connection: close`, not chunked, so chunked encoding is not the explanation here.

reddink:
**Windows runtime verified.** All three hosts are now covered, so the "compiled but never executed" caveat is closed.

Guix cross build at `a212fc842505`, `reddcoin-<ver>-win64.zip` run on Windows 11 (build 26200, AMD64).

The binary really does carry the new code and link:

```
imports CRYPT32.dll : True
has winstore URI    : True
```

```
> reddcoin-cli.exe -regtest checkupdates
{
  "localversion": "4.22.9.0",
  "remoteversion": "4.22.9.4",
  "updateavailable": true,
  "message": "Please download the latest version (4.22.9.4) from our official website",
  "errors": ""
}
```

Five consecutive calls, all `remote=[4.22.9.4] err=[]`.

A populated `remoteversion` with empty `errors` means `SSL_CTX_load_verify_store(ctx, "org.openssl.winstore://")` returned real anchors from the Windows ROOT store, `verify_peer` was active, and `api.github.com` chained to a trusted root and matched the hostname. Anchor failure is fatal in this code path, so there is no way to reach a populated version without verification having succeeded.

This also confirms the two build-system changes end to end on a real cross build: dropping `no-winstore` in depends produces a libcrypto whose winstore loader actually works at runtime, not merely one that links, and `AC_CHECK_LIB([crypt32], [CertOpenStore])` puts crypt32 where OpenSSL needs it.

## Verification status, all hosts

| Host | Builds | Links | Runtime `checkupdates` |
| --- | --- | --- | --- |
| Linux x86_64 | yes | yes | verified, `remoteversion` populated, `errors` empty |
| Windows x86_64 | yes | yes, CRYPT32.dll | verified, 5/5 calls |
| macOS x86_64 | yes | yes, Security + CoreFoundation | verified, 7/8 calls |

Plus the harness results already in the description: expired, wrong-host, self-signed and untrusted-root certificates all rejected, and an unreadable trust store fails closed rather than connecting unverified.

The single macOS outlier is the truncated-body case described in the previous comment. It is in the response handling this PR does not touch, is already recorded as blocker B3 in the design doc, and did not reproduce across 7 further calls.

RED-98 open question 2, "CA trust store on Windows and macOS", is answered: platform trust stores work on both, with no bundled CA list and no refresh obligation.

reddink:
**Correction to my earlier comment: the macOS empty-response case does reproduce.** I previously reported it as a one-off that did not recur. With a test aimed at the right condition it reproduces reliably enough to characterise.

## What reproduces

The failure is specific to **the first `checkupdates` after a cold daemon start**, on macOS only.

| Binary | Platform | Condition | Result |
| --- | --- | --- | --- |
| this PR (`a212fc842505`) | macOS 10.15.7 | warm daemon, 20 calls | 20 ok, 0 empty |
| this PR (`a212fc842505`) | macOS 10.15.7 | cold start, first call | 4 ok, **1 empty** (of 5 valid) |
| this PR, same source | Linux x86_64 | cold start, first call | 8 ok, 0 empty |
| 4.22.9 release | macOS 10.15.7 | cold start, first call | 7 ok, 0 empty |

## This is not the TLS change failing

Empty `errors` rules that out. Every failure in the fetch path throws, and `CheckForUpdates` turns a throw into `errors`: CA load failure, handshake failure, non-200 status, malformed status line, missing header terminator, and timeout. An empty `errors` field means none of those happened, so the handshake completed and a well-formed HTTP 200 with headers came back. Only `obj_response.read()` failed to find `tag_name`, leaving every field at its initial value.

The fault is downstream of anything this PR touches, in response handling.

The most likely mechanism remains the read loop silently accepting a short body: `async_read` with `transfer_all()` treats both `eof` and `stream_truncated` as success, so a connection that ends mid-body yields a truncated response that parses to nothing and reports success. `curl` from the same host confirms GitHub serves this endpoint with `Content-Length` and `connection: close`, not chunked, so chunked framing is not the explanation.

## What is not yet settled

Whether this PR *triggers* a latent fault or the fault is independent of it. The 4.22.9 control is a clean run, but it exercises completely different code: the old synchronous fetch in `rpc/server.cpp`, not `HttpsFetcher`. `HttpsFetcher` arrived with the OpenSSL 3.5.7 / `node::CheckForUpdates` refactor, before this PR.

Settling it needs a darwin build of `ddeacf898a^`, which is a guix cross build away. Worth doing before merge if reviewers want attribution nailed down; the alternative is to fix the underlying robustness gap, which makes the question moot.

## Suggested fix, separate from this PR

Treat a short body as an error rather than success: parse `Content-Length` and require the body to match it, and stop treating `stream_truncated` as a clean end when the response is incomplete. That converts a silent wrong answer into a reported one. This is blocker B3 territory in `doc/design/assisted-upgrade.md`, and it matters more than it looks: the same reader is slated to fetch installers, where silently accepting a truncated download would be considerably worse than a missing update notice.

## Incidental finding

The 4.22.9 release binary reports `"remoteversion":"22.9.4"` for tag `v4.22.9.4`. That is the unanchored-regex bug the anchored pattern in `node/update_check.cpp` already fixes, confirmed here against a real shipped binary rather than in theory.

reddink:
**Fixed the truncated-body issue** in `83e6e48263`, with tests.

## The bug

The response was read to the end of the stream, and both a clean end of file and a stream that ended without a TLS shutdown were flattened into success:

```cpp
m_ec = (ec == boost::asio::error::eof || ec == boost::asio::ssl::error::stream_truncated)
           ? boost::system::error_code{} : ec;
```

At that layer a connection that died mid-body is indistinguishable from one that finished, so a truncated body was returned as a complete one. Nothing downstream noticed: the status line and headers parse, the JSON parse of the short body quietly fails, and `CheckForUpdates` leaves every field at its initial value and reports no error. The caller gets a well formed answer saying no update exists.

## The fix

Response parsing moves into `node::ExtractHttpBody`, which accepts a body only when it can be shown to be whole:

- length matches `Content-Length`, or
- for a response delimited by the connection closing, the stream ended with a clean shutdown.

Which of the two endings arrived is now carried through rather than discarded. A chunked response is rejected outright rather than returned with its chunk framing embedded, since that framing is not decoded here. Header lookup is case insensitive per RFC 7230 and matches field names line by line, so a name appearing inside another header's value cannot be mistaken for the header itself.

## Tests

New `src/test/update_check_tests.cpp`, 10 cases: complete body, short body, overlong body, unusable `Content-Length`, connection-close framing with and without a clean shutdown, chunked, case-insensitive header matching, a header name embedded in another value, malformed responses, and non-200 statuses.

They have teeth. With the completeness checks removed, **7 of the 10 fail**; with them, all 10 pass. That was verified by temporarily neutering the checks and rebuilding, not assumed.

Live fetches against api.github.com still succeed across 6 cold starts on Linux, so no regression on the happy path.

## Not backported to 4.22.9

Deliberate. That branch runs the older synchronous fetch in `rpc/server.cpp`, where only a clean `eof` breaks the read loop and any other error, `stream_truncated` included, sets `errors`:

```cpp
if (error == boost::asio::error::eof) break;
if (error) { errors = error.message() + " " + std::to_string(error.value()); break; }
```

So it reports a truncated stream rather than silently swallowing it. That matches the control run: the 4.22.9 macOS binary produced 0 empty results across 7 cold starts, against roughly 1 in 5 for `HttpsFetcher`. Porting `ExtractHttpBody` there would mean restructuring code that does not have this defect, against the policy of keeping `v4.22.9-regtest` close to the release.

## Still open in this area

`HttpsFetcher` buffers the whole body in memory with no size cap, so a hostile or broken server can make the client allocate without bound. Left alone here to keep this change to the defect at hand; it belongs with the streaming rewrite in blocker B3.

reddink:
**Retested on macOS with the fix built in** (guix cross build at `83e6e4826336`, macOS 10.15.7).

The binary carries the change (all three new error strings present) and Security.framework is still linked.

## Result

| Run | ok | silent empty | truncation caught | harness artifact |
| --- | --- | --- | --- | --- |
| batch 1, 14 cold starts | 13 | **0** | 0 | 1 |
| batch 2, 14 cold starts | 13 | **0** | 0 | 1 |

26 of 26 valid cold starts returned `remoteversion 4.22.9.4` with empty `errors`. The silent empty result, previously about one cold start in five, did not occur once. The single artifact in each batch is my restart loop racing the RPC port, not a client failure.

## Being precise about what this does and does not show

**It does not show the new error path firing in the wild.** No truncation occurred in 28 attempts, so `Incomplete response` was never triggered. The fix converts a silent truncation into a reported error; it does not prevent truncation, so a run with no truncation cannot demonstrate it catching one. That behaviour is covered by the unit tests instead, where removing the completeness checks fails 7 of the 10 cases.

What this run does establish is no regression on the happy path, and that the silent failure mode is absent.

## Likely reason truncation stopped reproducing

An earlier batch in this session returned HTTP 403 on 8 of 12 cold starts. That was GitHub's unauthenticated API limit, exhausted by my own testing:

```
x-ratelimit-limit: 60
x-ratelimit-remaining: 0
```

Those 403s were reported correctly through `errors`, which incidentally exercises the non-200 path. After waiting for the quota to reset I kept both batches inside budget, finishing with 34 requests remaining, and saw no truncation at all.

The original truncations were observed when I had exhausted or nearly exhausted that limit. So the plausible reading is that truncated responses correlate with rate-limit pressure, GitHub cutting connections mid-response rather than always returning a clean 403. That is a hypothesis, not a demonstrated cause, and it does not change the case for the fix: whatever cuts the connection, a partial body must not be reported as a complete one.

It does mean a real user, making one update check per session rather than 28 in an afternoon, was unlikely to have hit this often.

reddink:
**Windows retest found a second cause, now fixed in `3a46b128b6`.**

## The retest

Guix cross build at `83e6e4826336` on Windows 11. The binary carries everything (`imports CRYPT32.dll`, `winstore URI`, `Incomplete response`, `without a clean shutdown` all present).

12 cold starts: **11 ok, 0 silent empty, 1 reported error**. The one error:

```
"errors": "Response uses an unsupported transfer encoding: chunked"
```

## What that means

api.github.com does not always answer with `Content-Length`. It switches to **chunked** intermittently, roughly 1 in 12 here. This corrects something I said earlier in this PR: I checked with `curl` and reported that the endpoint uses `Content-Length` and not chunked. That was true of every `curl` sample, including three more taken from the Windows box just now, but it is not true in general. GitHub varies, most likely on a cache miss.

It also means **the original macOS silent empties were probably chunked responses, not truncation**. A chunked body handed back with its chunk sizes embedded fails the JSON parse and leaves every field empty with no error, which is exactly the symptom. Truncation was my hypothesis; chunked is the mechanism actually observed. The completeness work still stands on its own, and the two share a cause: response framing was not being honoured.

## The fix

Rejecting chunked made the failure visible but still failed the check, so it is now decoded. Chunk extensions and trailers are skipped. Completeness is inherent to the decoder rather than checked afterwards: it returns only on the terminating zero length chunk, so a body that stops early throws instead of yielding what arrived. `Content-Length` is not consulted for a chunked response, since RFC 7230 forbids sending both and the terminating chunk is the proof of completeness. Any other transfer encoding is still rejected.

## Tests

`update_check_tests` is now 12 cases. New coverage: a single chunk, a body split across chunks, chunk extensions, trailers after the final chunk, an upper case header value, and five ways for a chunked body to be incomplete or malformed (no final chunk, cut mid-chunk, a declared size that overruns, a non-hex size, a missing chunk terminator).

The positive cases assert the decoded body equals the original, so they cannot pass unless decoding genuinely works.

Live fetches against api.github.com succeed across 5 cold starts on Linux.

## macOS retest, for completeness

Same build, macOS 10.15.7: 26 of 26 valid cold starts correct, 0 silent empties. No truncation or chunked response happened to occur in that window, so neither new path fired there; those are covered by the tests.

## Note for anyone reproducing this

GitHub's unauthenticated API limit is 60 requests per hour per IP, and cold-start testing burns through it quickly. Once exhausted every check returns HTTP 403, which is reported correctly but tells you nothing about the code. `curl -s https://api.github.com/rate_limit` is free to poll and worth checking before drawing conclusions from a run.

reddink:
**Retested both hosts on the chunked-decoding build** (guix cross build at `3a46b128b647`). Both binaries carry the decoder.

| Host | cold starts | ok | silent empty | chunk-related error | other |
| --- | --- | --- | --- | --- | --- |
| Windows 11 | 12 | 12 | 0 | 0 | 0 |
| macOS 10.15.7 | 12 | 12 | 0 | 0 | 0 |

The same 12-run test on the previous build gave 11 ok and 1 `Response uses an unsupported transfer encoding: chunked` on Windows. That error is now absent.

## What this does and does not prove

**It does not prove a chunked response was decoded in the field.** Once decoded, a chunked response is indistinguishable from a `Content-Length` one at the RPC surface: both return the version with empty `errors`. There is no observable signal for a successful decode, so absence of the error is the only thing measurable, and absence has two possible causes: the decoder worked, or no chunked response arrived.

I tried to pin down how often chunked actually occurs, replicating the client's exact request:

```
15 raw probes: chunked=0 content-length=15
```

Together with three earlier probes, that is 0 chunked in 18 samples, against 1 in 12 seen from the client. So chunked is rare and intermittent, and the honest reading of 24 clean runs is **consistent with the decoder working, but not evidence that it ran**.

What actually proves the decoding is `update_check_tests`: a single chunk, a body split across chunks, extensions, trailers, an upper case header value, and five malformed or incomplete variants. The positive cases assert the decoded body equals the original, so they cannot pass unless reassembly is correct.

What the field runs do establish is that the change causes no regression on the ordinary `Content-Length` path across two platforms, and that the silent empty result remains absent.

## Verification status

| Host | TLS verification | truncation guard | chunked decode |
| --- | --- | --- | --- |
| Linux | runtime verified | unit tested | unit tested, 5/5 live cold starts |
| Windows 11 | runtime verified | unit tested | unit tested, 12/12 live cold starts |
| macOS 10.15.7 | runtime verified | unit tested | unit tested, 12/12 live cold starts |

The TLS work is directly runtime-verified everywhere, since a populated `remoteversion` is unreachable without anchors loading and verification passing. The two HTTP framing fixes are proven by unit tests and corroborated, not proven, by the live runs.

PR #505, opened 2026-09-01
miner: use the template's parent when rewriting the coinbase height

reddink:
## Problem

The staker samples the chain tip before calling `CreateNewBlock()` and without holding `cs_main`, then hands that pointer to `IncrementExtraNonce()`, which rewrites the coinbase scriptSig with `pindexPrev->nHeight + 1`.

`CreateNewBlock()` re-reads the tip under `LOCK2(cs_main, m_mempool.cs)` and sets `hashPrevBlock` from what it finds there. When the tip advances between the two reads, the block is built on height H while its coinbase claims H, where `ContextualCheckBlock()` requires H + 1. The node then rejects a block it has just staked and signed:

```
Staker thread [125960328853056]: proof-of-stake block found 1bf42d7411e6f610...
ERROR: AcceptBlock: bad-cb-height, block height mismatch in coinbase
ERROR: ProcessNewBlock: AcceptBlock FAILED (bad-cb-height, block height mismatch in coinbase)
ERROR: ProcessNewBlock, block not accepted
```

Every occurrence is a stake reward silently discarded.

## Evidence

Two occurrences in ten days of `debug.log` on a mainnet staker, both with the coinbase height equal to the parent's height:

| date | staked block | coinbase claims | parent | parent height | should have claimed |
|---|---|---|---|---|---|
| 2026-08-25T03:44:34Z | `749a9523...` | 6,578,995 | `e78df6e1...` | 6,578,995 | 6,578,996 |
| 2026-09-01T14:23:10Z | `1bf42d74...` | 6,589,636 | `85ff6a4b...` | 6,589,636 | 6,589,637 |

Rate: 2 of 2311 staked blocks, about 0.09%. Both parents are in the main chain, so the blocks were otherwise valid and would have been accepted.

## Fix

Look the parent up from the template that `CreateNewBlock()` actually returned, using the `WITH_LOCK(::cs_main, return chainman->m_blockman.LookupBlockIndex(...))` idiom already present in `RegenerateCommitments()` in the same file.

The pre-call tip sample had no other use anywhere in the staker loop, so it is removed rather than left in place as a second source of truth.

Considered and rejected: skipping `IncrementExtraNonce()` for PoS blocks. There is no nonce grinding in PoS and `CreateNewBlock()` already sets a correct coinbase and merkle root, so it would work, but it changes the coinbase scriptSig format of every PoS block for no benefit.

## Testing

`g++ -fsyntax-only -std=c++17 -DHAVE_CONFIG_H` passes against this branch's own `bitcoin-config.h` and depends tree. The check was negative-controlled: renaming `LookupBlockIndex` to a bogus symbol makes it fail, so a clean exit means something.

Not runtime-verified. The race needs the tip to advance inside the window between the two tip reads, which surfaced twice in ten days on a busy mainnet staker, so it is not practical to reproduce on demand. Correctness here rests on the code reading.

The same change for the release line is in the v4.22.9-regtest PR.

YouTrack: https://reddink.youtrack.cloud/issue/RED-107

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -686,7 +686,6 @@ void PoSMiner(interfaces::StakingWallet& staking_wallet, ChainstateManager* chai
             // StakingWallet. No lock is held here, which is what this needs.
             staking_wallet.blockUntilSyncedToCurrentChain();
 
-            CBlockIndex* pindexPrev = chainman->ActiveChain().Tip();
             bool fPoSCancel = false;
             CScript scriptPubKey = GetScriptForDestination(dest);
             CBlock *pblock;
@@ -715,6 +714,17 @@ void PoSMiner(interfaces::StakingWallet& staking_wallet, ChainstateManager* chai
                 return;
             }
             pblock = &pblocktemplate->block;
+
+            // reddcoin: take the parent from the template we actually got, not
+            // from a tip sampled before the call. CreateNewBlock() re-reads the
+            // tip under cs_main and builds on whatever it finds there, so an
+            // earlier sample can already be one block behind by the time it
+            // returns. IncrementExtraNonce() rewrites the coinbase height from
+            // whichever index it is handed, and a stale one leaves a block whose
+            // coinbase claims its parent's height instead of its own. This node
+            // then rejects its own block as bad-cb-height and the stake is lost.
+            CBlockIndex* pindexPrev = WITH_LOCK(::cs_main, return chainman->m_blockman.LookupBlockIndex(pblock->hashPrevBlock));
+            assert(pindexPrev != nullptr);
             IncrementExtraNonce(pblock, pindexPrev, nExtraNonce);
 
             staking_wallet.notifyStakingStatusChanged();


reddink:
Investigated the 4 CI failures on [run 33576155692](https://github.com/reddcoin-project/reddcoin/actions/runs/33576155692). **None of them are caused by this change.**

## Why this change can only affect one of them

This PR touches `PoSMiner()` in `src/miner.cpp` and nothing else. `PoSMiner` has exactly one caller:

```
src/staker.cpp:308:  PoSMiner(*staking_wallet, chainman, connman, mempool, thread_id, running, interrupt);
```

That is the staking thread, started by `setstaking` / `-stake`. Unit tests never reach it, and neither does block generation in functional tests: `rpc/mining.cpp` has its own PoS block path that mirrors the logic rather than calling it. So the only failure that could plausibly be attributable is the one in a test that starts real staking threads, `rpc_staking.py`.

## 1. `32-bit + dash, gui [CentOS 9]` - `rpc_staking.py`, 1203 s

`test_unloading_a_wallet_stops_its_staking_thread` hung: `unloadwallet` never returned and hit the 1200 s RPC timeout.

**Not this PR.** The identical failure occurs on `node/tls-verify-update-check`, which does not touch `miner.cpp`, in [run 33567696385](https://github.com/reddcoin-project/reddcoin/actions/runs/33567696385):

```
rpc_staking.py failed, Duration: 1202 s
  rpc_staking.py, line 304, in run_test
    self.test_unloading_a_wallet_stops_its_staking_thread()
  rpc_staking.py, line 287, in test_unloading_a_wallet_stops_its_staking_thread
    wallet.unloadwallet()
'unloadwallet' RPC took longer than 1200.000000 seconds
```

Same file, same lines, same function, same timeout, and on x86_64 rather than i686, so it is not architecture specific either.

**This is a real bug on develop, not a slow test.** 1200 s is a hang. The node log shows the race window:

```
00:53:12.475 [httpworker.3] StakeWalletAdd Launching wallet.. [unload_staker]
00:53:12.476 [httpworker.0] ThreadRPCServer method=unloadwallet
00:53:12.477 [staker]       staker thread start
00:53:12.477 [staker-...]   Set proof-of-stake timeout: 500ms
            ... 20 minutes of nothing ...
01:13:12.629 [init]         Shutdown: In progress...
01:13:12.629 [staker-...]   staker thread exit
```

`unloadwallet` arrives 1 ms after `setstaking(True)` launches the thread, and lands while the staker is still starting. The staker then logs nothing until shutdown kills it, so its stop request appears to be missed when it races thread startup. That looks like a remaining gap in the RED-104 / RED-103 lifecycle work rather than anything new here, and it deserves tracking on its own.

## 2. `previous releases, qt5, DEBUG` - `validation_block_tests`

```
processnewblock_signals_ordering: block 6a3cb97b... unexpectedly failed, nStatus=43
```

A unit test, so `PoSMiner` is unreachable by construction. The test also documents this exact race itself:

> the screen above and this check are separate `cs_main` scopes with an unlocked `ProcessNewBlock` between them, and ten threads run this loop, so a concurrent thread can still invalidate an ancestor in that window

It passed on the sibling PR #507 and on the TLS branch at the same time.

## 3. `multiprocess, DEBUG` - `feature_bip68_sequence.py`

```
substring:     'non-BIP68-final'
error message: 'bad-txns-vout-negative'
```

The transaction the test built had a negative output, so it was rejected before sequence locks were ever evaluated. That is an input-amount problem in the test, the same class as the PoS UTXO-value adaptations elsewhere in the suite, and it is value dependent so it comes and goes. Nothing to do with the staking thread.

## 4. `TSan, depends, no gui`

```
ThreadSanitizer: CHECK failed: sanitizer_common.h:494 "((i)) < ((size_))" (0xffffffff, 0x10)
```

A sanitizer internal failure, not an application report. It also failed on PR #507 and on two `node/tls-verify-update-check` runs.

## These four jobs are flaky generally

`node/tls-verify-update-check` touches no consensus, mining or wallet code, yet hits all four job types across three runs, with a different test failing each time:

| Run | 32-bit CentOS 9 | multiprocess | previous releases | TSan |
| --- | --- | --- | --- | --- |
| 33567696385 | fail (`p2p_permissions`) | fail (`mempool_unbroadcast`) | fail (`rpc_staking`) | pass |
| 33586270580 | fail | cancelled | pass | fail |
| 33589508947 | pass | pass | pass | fail |

Different tests failing in the same jobs run to run is the signature of load-related flakiness rather than a code defect, with the `rpc_staking` hang the one genuine bug among them.

## Suggested action

Re-run the failed jobs. Nothing here blocks this PR. The `unloadwallet` hang should be filed separately against develop, since it is a real deadlock that will keep costing CI runs 20 minutes at a time until it is fixed.

reddink:
Filed the `unloadwallet` hang separately as **RED-109**: https://reddink.youtrack.cloud/issue/RED-109

Root cause, for the record: `CStakeman::StakeWalletAdd` creates the staking thread, and that thread only subscribes to the wallet's unload notification once it is already running, at `src/staker.cpp:302`. Everything between `std::thread` construction and that line is a window where the thread holds a reference to the wallet but nothing is listening for its unload. An `unloadwallet` landing in that window fires `NotifyUnload` with no handler attached, so the interrupt is never delivered, the staker settles into its loop still holding the wallet, and `UnloadWallet()` waits forever for a reference that will not be released.

The node log confirms the interrupt never arrived: `ThreadStaker` logs "Staking thread stopping, wallet unloaded" when the handler fires, and that line is absent for the whole 20 minutes.

Nothing to do on this PR. Re-running the failed jobs should be enough.

PR #509, opened 2026-09-02
doc: sign the SHA256SUMS manifest that actually gets published

reddink:
Resolves blocker B1 of RED-98 on the process side: `SHA256SUMS.asc` had never been published.

## The gap

`SHA256SUMS.asc` returned 404 on `download.reddcoin.com` for **every** release checked - 4.22.8, 4.22.9, 4.22.9.3 and 4.22.9.4. The manifest binding release artifacts to their hashes was authenticated by nothing but TLS to the web host.

Nothing was missing to make it work. The release key existed, the manifest content was correct, and `guix-attest` already produces signatures. The process simply routed through an upstream step gated on *"After 3 or more people have guix-built and their results match"*:

```bash
cat "$VERSION"/*/all.SHA256SUMS.asc > SHA256SUMS.asc
```

Reddcoin releases are signed by one release manager, so that gate is never met and the step was never reached.

## Two ways the obvious fix fails

**1. `all.SHA256SUMS` is not the published file.** `guix-attest` writes it over every fragment, including `reddcoin-<version>-codesignatures-<version>.tar.gz`, an intermediate of the codesigning step that is not uploaded:

| file (4.22.9.4) | sha256 | entries |
| --- | --- | --- |
| `all.SHA256SUMS` | `ec9074c6...` | 23 |
| published `SHA256SUMS` | `b60e66fb...` | 22 |

A detached signature over one reports BAD against the other, which is indistinguishable from tampering. The difference is not stable across releases either: 4.22.9 shipped without macOS codesigning so its `all.SHA256SUMS` matched what was published, while 4.22.9.4's did not.

**2. The manifest listed the `*-debug*` archives**, which the upload step excludes on purpose. For 4.22.9.4 that left 7 of 22 entries pointing at 404s, so the obvious verification command failed on every one of them and only came out clean with `--ignore-missing`. A verifier whose check reports failures cannot tell a deliberate omission from a corrupted download.

## Where the filtering belongs

Nothing under `contrib/guix` strips the debug archives, which is an easy thing to assume it does. `libexec/build.sh:462` builds each fragment from `find "$ACTUAL_OUTDIR" -type f`, and `guix-attest` only concatenates fragments.

That is correct for `guix-attest`. `all.SHA256SUMS` and `noncodesigned.SHA256SUMS` are the guix.sigs attestations, where covering every artifact built is the point, since it is what lets two builders compare results. The published `SHA256SUMS` is a different artifact for a different audience, so the narrowing happens at publish time. The doc says so at the point it happens, and notes the exclusion must stay in step with the upload command's `-not -name "*debug*"`, since the two are written separately and will otherwise drift.

## Changes

- Derive the manifest from what is actually uploaded, and sign that. Both files must be uploaded together whenever either is regenerated, since a signature beside a stale manifest verifies BAD.
- Add **Verify the published release**: fetch manifest and signature from the server, verify, then download every artifact named and run `sha256sum -c`. This catches a manifest listing unpublished files and a signature regenerated without its manifest, neither visible from the build directory. Includes the throwaway-keyring form, since a keyring that already trusts the release key cannot fail the way a new user's does.
- Add **Publishing the release key**, covering both keyservers. `keys.openpgp.org` publishes the user ID only after the owner answers a confirmation mail, and until then serves a key with no user ID that `gpg` refuses to import, so the upload is inert despite returning HTTP 200 with a well-formed key block. The check pipes through `gpg` rather than trusting a status code.
- Point the upload target at `download.reddcoin.com` instead of `bitcoincore.org`.
- Add the release key fingerprint to `contrib/builder-keys/keys.txt`.

## Verified

Against the live 4.22.9.4 release, from the server with the key fetched fresh from a keyserver into an empty keyring:

```
gpg: Good signature from "John Nash (reddink) <john@redd.ink>"
Primary key fingerprint: ABED C448 9B91 88E4 5C23  42A8 2E91 240B 293B A5D3

reddcoin-4.22.9.4-osx-signed.dmg:          OK
reddcoin-4.22.9.4-win64-setup-signed.exe:  OK
```

15 manifest entries, all downloadable, `sha256sum -c` clean without `--ignore-missing`.

## Earlier releases

4.22.8, 4.22.9 and 4.22.9.3 keep their unsigned manifests. Signing them retroactively was considered and deliberately not done: 4.22.9.4 is the current release, and the point of the update check is to move people onto it rather than to make older downloads more attractive to verify.

The process from here forward signs every release.

YouTrack: https://reddink.youtrack.cloud/issue/RED-98


reddink:
Amended to add a **Publishing the release key** section, since a signature is no use to anyone who cannot fetch the key that made it. Force-pushed over the previous commit; no review had happened yet.

It gives the upload command for both keyservers and says why both are worth having, because they behave differently:

- **keys.openpgp.org** validates the address and distributes only self-signatures, stripping third-party ones. It accepts a single user ID. Crucially it publishes identity information only after the owner answers a confirmation mail, so until that is done the key is retrievable **by fingerprint** but not by email or key ID. Fingerprint is the lookup that matters here, since that is what `contrib/builder-keys/keys.txt` distributes, but the distinction is worth knowing before concluding an upload failed.
- **keyserver.ubuntu.com** performs no validation and accepts whatever it is given.

The verification step pipes the response through `gpg --show-keys` rather than checking the HTTP status:

```bash
curl -sf "https://keys.openpgp.org/vks/v1/by-fingerprint/${SIGNER_KEY}" | gpg --show-keys
curl -sf "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x${SIGNER_KEY}" | gpg --show-keys
```

That is not pedantry. I first reported the key as present on keyserver.ubuntu.com purely on a 200 response, which is not evidence: a lookup that found nothing can still answer with a page that is not a key. Importing what it actually serves into a throwaway keyring confirmed the fingerprint and user ID matched, but the status code alone would not have.

Current state of the two, checked just now:

| Keyserver | Serving the release key |
| --- | --- |
| keyserver.ubuntu.com | yes, fingerprint and UID verified by import |
| keys.openpgp.org | no, 404 |

Also added an armoured export to keep on the download server, so verification does not depend on a keyserver being reachable at all.

One structural fix while in there: the new section had left the upload steps sitting underneath it, which read as though uploading the binaries were part of publishing the key. They now sit under their own **Publish the release** heading.

reddink:
The release key is now uploaded to keys.openpgp.org, and testing it against the live server turned up a correction to what this PR said, fixed in `c662516a3d`.

**The key is there**, and a by-fingerprint lookup returns it:

```
pub   rsa3072 2024-02-14 [SCEA]
      ABEDC4489B9188E45C2342A82E91240B293BA5D3
sub   rsa3072 2024-02-14 [E]
```

**But an unconfirmed upload is of no practical use**, which is stronger than what the doc claimed. The earlier wording said that until the address is confirmed the key "can still be fetched by fingerprint, which is the lookup that matters here". That is wrong in a way that would mislead anyone relying on it: the key comes back with no user ID attached, and gpg refuses to import a key in that state:

```
gpg: Total number processed: 1
gpg:           w/o user IDs: 1
```

Nothing lands in the verifier's keyring. A by-email lookup returns 404. The response is 1821 bytes against 2484 from keyserver.ubuntu.com, the difference being the stripped user ID and third-party signatures.

So answering the confirmation mail is required, not a nicety, and the section now says so.

Current state of the two:

| Keyserver | Serves the key | Serves the user ID | Importable |
| --- | --- | --- | --- |
| keyserver.ubuntu.com | yes | yes | yes |
| keys.openpgp.org | yes | not until confirmed | not yet |

Worth noting how this was caught, since it is the same trap the verification step in this PR already warns about one level up: the lookup returns HTTP 200 and a well formed PGP key block, so any check short of importing it would have called this done.

reddink:
**No, nothing under `contrib/guix` strips the debug archives** - worth stating in the doc, since it is an easy and reasonable thing to assume.

`libexec/build.sh:462` builds each fragment from every file in the output directory:

```sh
find "$ACTUAL_OUTDIR" -type f
```

and `guix-attest` only concatenates those fragments. So the debug archives are present from `SHA256SUMS.part` all the way through `all.SHA256SUMS`, and nothing removes them at any point.

That is correct behaviour for `guix-attest`. `all.SHA256SUMS` and `noncodesigned.SHA256SUMS` are the guix.sigs attestations, where covering every artifact that was built is exactly the point: it is what lets two builders compare results. The published `SHA256SUMS` is a different artifact for a different audience, so the narrowing belongs at publish time rather than in the attestation.

Added in `3e69f881be`:

- A note at the filtering step explaining that no guix script does this, and that the exclusion has to stay in step with the upload command's `-not -name "*debug*"`. The two are written out separately and will otherwise drift, which is how the mismatch arose in the first place.
- A **Verify the published release** step that checks the upload from the server rather than the build tree: fetch manifest and signature over the network, verify the signature against them, then download every artifact the manifest names and run `sha256sum -c`. That catches a manifest listing files that were never uploaded, and a signature regenerated without its manifest. Neither is visible from the build directory, and both happened during this work.
- The throwaway-keyring variant, since verifying with a keyring that already trusts the release key cannot fail the way a new user's does. The uncertified-key warning that comes with a fresh keyring is called out as expected, so it is not mistaken for a failure.

`sha256sum -c` is documented as having to pass without `--ignore-missing`. Needing that flag means the manifest lists something that was not uploaded, and a verifier who sees failures cannot tell a deliberate omission from a corrupted download.

Confirmed against the live 4.22.9.4 release: 15 entries, all downloadable, good signature, `sha256sum -c` clean.

PR #510, opened 2026-09-03
build: export the host triplet to C++ as HOST_TRIPLET

reddink:
Closes blocker **B4** of the assisted-upgrade design. First commit of RED-110 (phase 1).

## Why

`configure.ac` derives `TARGET_OS` for automake conditionals but never makes the autoconf host triplet available to the code, so nothing can tell at runtime which platform's build it is.

The update check needs exactly that. guix names every release artifact after the triplet it was built for (`contrib/guix/libexec/build.sh:406,431`), so the triplet is what maps a running binary to the file a user should download.

`TARGET_OS` is too coarse for this. It cannot distinguish `x86_64-linux-gnu` from `aarch64-linux-gnu`, and those are separate published artifacts.

## What

`AC_CANONICAL_HOST` already computes `$host` at the top of `configure.ac`, so this is a define placed next to it rather than any new detection:

```m4
AC_DEFINE_UNQUOTED([HOST_TRIPLET], ["$host"], [The autoconf host triplet this build targets])
```

## Verified

End to end on `x86_64-pc-linux-gnu`, rather than just checking the macro is present:

- the autoheader stanza reaches `src/config/bitcoin-config.h.in`
- a full `configure` run produces `#define HOST_TRIPLET "x86_64-pc-linux-gnu"`
- a program including that header compiles and prints the value

Nothing consumes it yet. The artifact mapping that uses it is the next commit of RED-110, kept separate so this stands on its own.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,15 @@ fi
 
 AC_CANONICAL_HOST
 
+dnl The autoconf host triplet this build targets, exported to C++.
+dnl
+dnl The update check needs it to name the release artifact built for this
+dnl platform: guix names every output after the triplet, so the triplet is what
+dnl maps a running binary to its download. TARGET_OS, derived further down, is
+dnl too coarse for that. It cannot tell x86_64 Linux from aarch64 Linux, and the
+dnl published artifacts differ per triplet.
+AC_DEFINE_UNQUOTED([HOST_TRIPLET], ["$host"], [The autoconf host triplet this build targets])
+
 AH_TOP([#ifndef BITCOIN_CONFIG_H])
 AH_TOP([#define BITCOIN_CONFIG_H])
 AH_BOTTOM([#endif //BITCOIN_CONFIG_H])


reddink:
Added the artifact mapping, `324369a313`. Still nothing user-visible; the RPC fields and the GUI change follow.

## A trap worth flagging for review

The host triplet **cannot be used verbatim** to build a filename. The artifacts are named after the three-field alias guix is invoked with, while `HOST_TRIPLET` records autoconf's `$host`, which `config.sub` has already expanded to four fields. They agree for Windows and macOS and differ for **every** Linux host:

| guix `HOST` (artifact name) | `$host` (`HOST_TRIPLET`) |
| --- | --- |
| `x86_64-linux-gnu` | `x86_64-pc-linux-gnu` |
| `arm-linux-gnueabihf` | `arm-unknown-linux-gnueabihf` |
| `aarch64-linux-gnu` | `aarch64-unknown-linux-gnu` |
| `riscv64-linux-gnu` | `riscv64-unknown-linux-gnu` |
| `powerpc64-linux-gnu` | `powerpc64-unknown-linux-gnu` |
| `powerpc64le-linux-gnu` | `powerpc64le-unknown-linux-gnu` |
| `x86_64-w64-mingw32` | same |
| `x86_64-apple-darwin18` | same |

guix never passes `--host`; it relies on `CONFIG_SITE` alone, and `depends/config.site.in` sets `host_alias="@HOST@"`, which autoconf then canonicalises. So the vendor field appears in `$host` and not in the filename.

Keying on `$host` without dropping the vendor would name `reddcoin-<version>-x86_64-pc-linux-gnu.tar.gz`, which does not exist. Both forms are accepted, since a build from source legitimately produces the canonical one while a release build produces the alias.

## Design note

The mapping returns **both** the GUI and daemon artifact rather than taking an `is_gui` argument. This code is shared by both binaries, and `checkupdates` is reachable from `reddcoin-cli` *and* the Qt console, so a single "am I a GUI" flag set at startup would be wrong whenever the RPC is called from the other surface. The presenter chooses.

On Linux the two are equal: one tarball carries both binaries and no separate installer is published.

Apple Silicon needs no special case. `HOST_TRIPLET` is fixed at build time and no arm64 Darwin build is published, so a Mac running this binary is by definition running the x86_64 one and already reports `x86_64-apple-darwin18`. It is offered that build, which is the one that runs under Rosetta.

## Tests

`release_artifacts_tests`, 9 cases, table driven over every host in the guix `HOSTS` list. Each expected filename is taken from the **published 4.22.9.4 `SHA256SUMS`**, so a change to the guix naming scheme fails the tests rather than being silently encoded as a stale convention.

Verified the tests have teeth rather than assuming it: removing the normalisation fails **24 assertions**, naming exactly the file that would 404.

```
reddcoin-4.22.9.4-powerpc64-unknown-linux-gnu.tar.gz
   != reddcoin-4.22.9.4-powerpc64-linux-gnu.tar.gz
```

PR #512, opened 2026-09-03
node: map the host triplet to its published release artifacts

reddink:
Second commit of RED-110 (assisted upgrade, phase 1), rebased onto develop now that #510 has merged. Builds on the `HOST_TRIPLET` define that PR added.

Nothing user-visible yet. The `checkupdates` RPC fields and the GUI change follow.

## What

Maps a host triplet to the release artifacts published for it. A pure function, no I/O, no network:

```
x86_64-w64-mingw32     -> -win64-setup-signed.exe / -win64.zip
x86_64-apple-darwin18  -> -osx-signed.dmg         / -osx64.tar.gz
<arch>-linux-<abi>     -> -<triplet>.tar.gz       (both)
```

## The trap this exists to avoid

The host triplet **cannot be used verbatim** to build a filename. The artifacts are named after the three-field alias guix is invoked with, while `HOST_TRIPLET` records autoconf's `$host`, which `config.sub` has already expanded to four fields. They agree for Windows and macOS and differ for **every** Linux host:

| guix `HOST` (artifact name) | `$host` (`HOST_TRIPLET`) |
| --- | --- |
| `x86_64-linux-gnu` | `x86_64-pc-linux-gnu` |
| `arm-linux-gnueabihf` | `arm-unknown-linux-gnueabihf` |
| `aarch64-linux-gnu` | `aarch64-unknown-linux-gnu` |
| `riscv64-linux-gnu` | `riscv64-unknown-linux-gnu` |
| `powerpc64-linux-gnu` | `powerpc64-unknown-linux-gnu` |
| `powerpc64le-linux-gnu` | `powerpc64le-unknown-linux-gnu` |
| `x86_64-w64-mingw32` | same |
| `x86_64-apple-darwin18` | same |

guix never passes `--host`; it relies on `CONFIG_SITE` alone, and `depends/config.site.in` sets `host_alias="@HOST@"`, which autoconf then canonicalises. So the vendor field appears in `$host` and not in the filename.

Keying on `$host` without dropping the vendor would name `reddcoin-<version>-x86_64-pc-linux-gnu.tar.gz`, which does not exist. Both forms are accepted, since a build from source legitimately produces the canonical one while a release build produces the alias.

## Why it returns both artifacts rather than taking an is_gui flag

The choice of file is a function of `(host, is_gui)`, not host alone: a `reddcoin-qt` user wants the installer or the disk image, a `reddcoind` user wants the archive.

This code is shared by both binaries, and `checkupdates` is reachable from `reddcoin-cli` **and** from the Qt console, so a single "am I a GUI" flag set at startup would be wrong whenever the RPC is called from the other surface. Both are returned and the presenter chooses.

On Linux the two are equal: one tarball carries both binaries and no separate installer is published.

## Apple Silicon needs no special case

`HOST_TRIPLET` is fixed at build time and no arm64 Darwin build is published, so a Mac running this binary is by definition running the x86_64 one and already reports `x86_64-apple-darwin18`. It is offered that build, which is the one that runs under Rosetta. Naming it as the Intel build is left to the presentation layer.

An unpublished host resolves to nothing rather than to a guess, because naming a file that will 404 is worse than saying nothing.

## Tests

`release_artifacts_tests`, 9 cases, table driven over every host in the guix `HOSTS` list. Each expected filename is taken from the **published 4.22.9.4 `SHA256SUMS`**, so a change to the guix naming scheme fails the tests rather than being silently encoded as a stale convention.

Verified the tests have teeth rather than assuming it. Removing the normalisation fails **24 assertions**, naming exactly the file that would 404:

```
reddcoin-4.22.9.4-powerpc64-unknown-linux-gnu.tar.gz
   != reddcoin-4.22.9.4-powerpc64-linux-gnu.tar.gz
```

Rebuilt and re-run after the rebase onto develop, not just before it.

## Note on the branch name

`node/host-triplet-define` now carries only the mapping; the define it was named for merged in #510. Kept as is to avoid churn, but happy to move it to something like `node/release-artifact-mapping` if you would rather the name matched.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -188,6 +188,7 @@ BITCOIN_CORE_H = \
   node/coin.h \
   node/coinstats.h \
   node/ca_store.h \
+  node/release_artifacts.h \
   node/context.h \
   node/psbt.h \
   node/transaction.h \
@@ -367,6 +368,7 @@ libbitcoin_server_a_SOURCES = \
   node/coin.cpp \
   node/coinstats.cpp \
   node/ca_store.cpp \
+  node/release_artifacts.cpp \
   node/context.cpp \
   node/interfaces.cpp \
   node/psbt.cpp \
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -147,6 +147,7 @@ BITCOIN_TESTS =\
   test/txrequest_tests.cpp \
   test/txvalidation_tests.cpp \
   test/uint256_tests.cpp \
+  test/release_artifacts_tests.cpp \
   test/update_check_tests.cpp \
   test/uint512_tests.cpp \
   test/util_tests.cpp \
--- a/src/node/release_artifacts.cpp
+++ b/src/node/release_artifacts.cpp
@@ -1,0 +1,117 @@
+// Copyright (c) 2014-2026 The Reddcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#if defined(HAVE_CONFIG_H)
+#include <config/bitcoin-config.h>
+#endif
+
+#include <node/release_artifacts.h>
+
+#include <string>
+#include <vector>
+
+namespace {
+//! Split a host triplet on '-'.
+std::vector<std::string> SplitTriplet(const std::string& host_triplet)
+{
+    std::vector<std::string> fields;
+    std::string::size_type start{0};
+    while (true) {
+        const std::string::size_type dash{host_triplet.find('-', start)};
+        if (dash == std::string::npos) {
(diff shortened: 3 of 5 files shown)


reddink:
Closed by a branch rename, not abandoned. Continues as #513, same commit `642bdd84be` and the same diff.

The branch was renamed from `node/host-triplet-define` to `node/release-artifact-mapping` to match its contents, since the define it was named for merged separately in #510.

Worth recording for anyone who hits this: GitHub's branch rename API retargets open pull requests **within the same repository**, but this is a cross-repository PR from a fork, so renaming the head branch closed it instead. A closed PR whose head ref no longer exists cannot be reopened either. Renaming the head branch of a fork PR therefore always costs a new PR.

PR #515, opened 2026-09-03
rpc: name the exact artifact this host needs in checkupdates

reddink:
Third commit of RED-110 (assisted upgrade, phase 1), on top of the mapping merged in #513.

## What

`checkupdates` could say a newer version exists and point at the directory holding it, but not say which of the **fifteen** files in that directory the machine asking actually needs. Six fields close that gap:

| Field | |
| --- | --- |
| `hosttriplet` | the host triplet this build targets |
| `platform` | short platform name used in artifact filenames |
| `guiartifact` | the file a `reddcoin-qt` user should install |
| `guiartifactlink` | its direct download link |
| `daemonartifact` | the file a `reddcoind` user should install |
| `daemonartifactlink` | its direct download link |

Both surfaces are reported rather than the node guessing which the caller is. This RPC is reachable from `reddcoin-cli` **and** from the Qt console, so a flag set at startup would be wrong whenever it is consulted from the other one. On Linux the two are equal: one tarball carries both binaries.

Every field is present and empty rather than absent when unknown, so the shape of the result does not depend on whether the check succeeded. That matches how the existing fields behave.

## Verified against the live release, not only in tests

On `x86_64-pc-linux-gnu`:

```
"hosttriplet":     "x86_64-pc-linux-gnu",
"platform":        "x86_64-linux-gnu",
"guiartifact":     "reddcoin-4.22.9.4-x86_64-linux-gnu.tar.gz",
"guiartifactlink": "https://download.reddcoin.com/bin/reddcoin-core-4.22.9.4/reddcoin-4.22.9.4-x86_64-linux-gnu.tar.gz",
```

That link answers **HTTP 200**.

The link a mapping without the vendor normalisation would have produced answers **HTTP 404**:

```
.../reddcoin-4.22.9.4-x86_64-pc-linux-gnu.tar.gz     404
.../reddcoin-4.22.9.4-x86_64-linux-gnu.tar.gz        200
```

Note `hosttriplet` and `platform` legitimately differ in that output. The first is what configure recorded, the second is what the artifact is named after. That difference is the whole reason the mapping normalises, and it applies to every Linux host.

## Prereleases are deliberately not named

No prerelease has ever been published, so the naming convention for one is unverified. Naming a file that does not exist turns a working notice into a broken link, so the artifact fields stay empty for a prerelease and the existing directory link still applies.

## Also

`HostTriplet()` is added alongside the mapping so the single dependency on the generated config header stays in one translation unit rather than spreading to every caller that wants to report the triplet.

Nothing user-visible yet: the GUI change is the last commit of this phase.


reddink:
CI status on this PR: three jobs are red, **none of them caused by this change**.

| Job | Failure | Cause |
| --- | --- | --- |
| multiprocess | `rpc_staking.py`, 1202s `unloadwallet` hang | RED-109, filed separately |
| TSan | `ThreadSanitizer: CHECK failed`, exit 137 | known sanitizer/OOM flake |
| 32-bit CentOS 9 | depends build break configuring Qt | **RED-111**, filed just now |

The check that matters for this PR passes:

```
rpc_checkupdates.py    ✓ Passed
```

There is a functional test for this RPC and it passes with the six new fields, so the `RPCResult` declaration matches what the RPC actually returns. Nothing in any of the three logs mentions `Internal bug detected`, an `RPCResult` mismatch, or `checkupdates` failing.

The 32-bit failure is new and worth knowing about, since it will block every PR until fixed. It dies in depends before any project code compiles:

```
ERROR: Feature 'system-freetype' was enabled, but the pre-condition
       'features.freetype && libs.freetype' failed.
```

freetype and fontconfig both build and cache successfully in that same run, so Qt is failing to detect a library that is present rather than one that is missing. It was masked until now by a cached depends tree: the last passing run compiled Qt zero times, this one compiled it once. The cache key turned over when `depends/packages/openssl.mk` changed in #503, which removed the cache that was hiding the breakage rather than causing it.

Full analysis in https://reddink.youtrack.cloud/issue/RED-111

PR #523, opened 2026-09-03
contrib: sign the release manifest for the client

reddink:
Phase 2b of RED-98 (RED-113). Adds the signature the client will verify, alongside the `SHA256SUMS.asc` that people and third parties verify.

`SHA256SUMS.asc` proves a release to a person with OpenPGP tooling. It cannot prove it to the client, which has no OpenPGP and, on Windows, no `gpgv` either. Both signatures cover the same manifest and both get published.

The format this implements is now specified in **REP-1018**, merged into the reps repository as `docs/rep-1018.md`. The purpose value it derives under is recorded in **REP-0005**, the derivation path registry.

## The tool

`contrib/release-signing/sign-release-manifest.py`, four subcommands:

| | |
| --- | --- |
| `selftest` | run the official BIP340 test vectors |
| `pubkey` | print the x-only public key for a mnemonic |
| `sign` | sign a manifest |
| `verify` | check a signature, non-zero exit on failure |

It depends on nothing but Python and this repository. The BIP340 implementation and the official vectors already live in `test/functional/test_framework`, so this reuses them rather than carrying a second copy of signature code that nothing else exercises. That matters for an air-gapped box: no packages to install, and it can prove the implementation correct on that machine **before** it signs anything users will trust.

## Design points worth reviewing

**`sign` refuses to write a signature it cannot itself verify**, and prints the public key it used. A signature made with the wrong key verifies happily against itself and against nothing a released client accepts, so the printed key is the check that cannot be skipped.

**A mistyped mnemonic is rejected rather than signed with.** The words are checked against the BIP39 list and the phrase against its own checksum. Without this the only guard was an operator comparing 64 hex characters by eye, which fails worst during recovery onto a rebuilt machine, where one wrong word silently derives a perfectly usable key for a wallet that has never signed anything. The wordlist is parsed from `src/util/lang/bip39_english.h`, the header the client already ships, so there is one wordlist in the repository and it cannot drift from the one the wallet derives with.

**The manifest is read as bytes, never as text.** The signature covers the file exactly as it will be served; newline or encoding normalisation on the way through would sign something the server does not serve. Same class of trap as `all.SHA256SUMS` not being the published manifest.

**`verify` exits non-zero**, so an upload can be gated on it.

**The `.sig` carries only the signature, not the public key.** Including the key would invite verifying a signature against a key taken from the same file, which proves nothing. The client pins the key.

## What becomes permanent

All of it is fixed once a client ships that reads it, which is why it is specified in REP-1018 rather than left as constants in a script.

**The message** is `TaggedHash("Reddcoin/ReleaseManifest")` over the manifest bytes, matching `TaggedHash()` in `src/hash.h`. The release key sits on the same curve as transaction keys, so signing a bare `SHA256` would leave a signature that could be meaningful in another context.

**The derivation path** is `m/1018'/4'/0'`, deliberately not under the `m/44'/4'/...` wallet tree, so a release key cannot collide with a spending key.

It was `m/1017'` when this PR was opened, and that was wrong. `1017'` is lnd's: `BIP0043Purpose = 1017` in `keychain/derivation.go`, with the hierarchy `m/1017'/coinType'/keyFamily'/0/index` and `KeyFamilyMultiSig = 0`. No key would ever have collided, since the release seed derives nothing else, but the path described the release key as a Lightning multisig branch for coin type 4, and a derivation path is read by people. The replacement was checked against the BIP-allocated purposes, the SLIP-reserved 10001-19999 range, and known unregistered users.

The number matches the REP that specifies it, which is what BIP43 asks for: schemes should take the number of the document defining them, so a purpose value leads back to a specification rather than to a constant in someone's source tree. REP-1018 specifies purpose `1018'`.

This was free to change only because no production key exists yet. One step later it would have cost a client release, since a mnemonic does not identify a key without its path.

**The BIP32 HMAC key stays the literal `"Bitcoin seed"`.** It is fixed by BIP32 for every implementation and every coin, not a name to localise. Renaming it would derive a different key from the same words while looking perfectly correct, and would put the key beyond the reach of every standard BIP39 recovery tool, which matters because the threat model here is loss rather than compromise.

## The client and the tool are held to the same answer

`src/test/release_manifest_tests.cpp` is the reason the last commit exists. The signature is produced by Python on an air-gapped machine and checked by C++ in the client, and nothing forced the two to agree. A divergence would not have surfaced until every release failed to verify for every user.

It asserts REP-1018 section 5.2's published vectors against the C++ side: `TaggedHash` over the manifest, and the reference signature checked through the same `XOnlyPubKey::VerifySchnorr` call a client will make. The negative cases carry most of the value. A modified manifest, an appended line, a modified signature and a wrong key all fail, and so does a tag one character short or one character long, which is what shows the domain separation exists rather than being described in a comment.

Two traps worth naming for a reviewer, since both produce a test that passes while proving nothing:

- `uint256::ToString()` reverses byte order, so comparing it against the tool's hex tests a byte-swapped value. The test compares raw order.
- `Span`'s container constructors are lifetime-bound, so building one from a `ParseHex` temporary risks a dangling diagnostic that is fatal under `-Werror`. Every input in the test is a named local.

Deliberately not tested: `VerifySchnorr` asserts `sigbytes.size() == 64`, so a malformed `.sig` aborts the process rather than returning false. That length check is the caller's job and belongs in the client code REP-1018 section 3.6 step 2 requires. Exercising it here would kill the test binary, so there is a comment saying so instead.

## Verified

Against the live 4.22.9.4 manifest fetched from the server:

- all **15 official BIP340 vectors** pass
- sign then verify round trips, exit 0
- **five tamper cases** each exit 1: a manifest altered by one byte, a flipped signature byte, a different public key, a malformed short signature, and a manifest regenerated with an extra line
- the signature file is 129 bytes, one line, 128 lowercase hex plus a newline, as REP-1018 section 3.5 requires
- `gpg --verify` on the published `SHA256SUMS.asc` returns a good signature over the identical bytes, so the two signatures do cover the same file
- mnemonic validation rejects a bad checksum, a non-BIP39 word and a wrong length. Neutering the check makes the mistyped phrase derive a different key, which is what shows the check is load bearing rather than decorative

Exit codes checked directly rather than through a pipe, since a release script would branch on them.

## Note for phase 4

The published manifest lists signed and unsigned variants side by side, so artifact matching must be on the exact filename field with exactly one matching line, as REP-1018 section 3.6 step 6 requires. A substring match for `signed.exe` also matches `reddcoin-4.22.9.4-win64-setup-unsigned.exe`, which would hand a Windows user the unsigned installer while reporting a verified download.

## Not in this PR

The production key does not exist yet. `doc/release-process.md` has a placeholder for its public key, to be filled when it is generated on the air-gapped machine.

Client-side verification is phase 4 and consumes the format fixed here. It must not be written until a real `.sig` is published, or it will pass its own tests against synthetic data and fail in production, which is the trap B1 already posed once for the `.asc`.

YouTrack: https://reddink.youtrack.cloud/issue/RED-113

### DIFF
--- a/contrib/release-signing/sign-release-manifest.py
+++ b/contrib/release-signing/sign-release-manifest.py
@@ -81,6 +81,16 @@ def tagged_hash(tag, data):
     return hashlib.sha256(ss + ss + data).digest()
 
 
+# Where the client keeps the English BIP39 wordlist. It sits in src/util/lang on
+# the development line and in src/wallet on the 4.22.9 release line, so both are
+# tried: this tool is copied to an air-gapped machine and should be one file that
+# works from either checkout rather than two that can drift apart.
+WORDLIST_PATHS = (
+    ('src', 'util', 'lang', 'bip39_english.h'),
+    ('src', 'wallet', 'bip39_english.h'),
+)
+
+
 def load_wordlist():
     """The English BIP39 wordlist, read from the header the client uses.
 
@@ -88,8 +98,13 @@ def load_wordlist():
     second copy here could drift from the one the wallet derives with, and the
     divergence would only ever show up as a key that does not match.
     """
-    path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                        '..', '..', 'src', 'util', 'lang', 'bip39_english.h')
+    root = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..')
+    tried = [os.path.join(root, *parts) for parts in WORDLIST_PATHS]
+    found = [p for p in tried if os.path.exists(p)]
+    if not found:
+        raise SystemExit("no BIP39 wordlist found, looked in:\n  " + "\n  ".join(tried))
+
+    path = found[0]
     words = re.findall(r'^\s*"([a-z]+)",?\s*$', open(path, encoding='utf8').read(), re.MULTILINE)
     if len(words) != 2048:
         raise SystemExit("expected 2048 words in {}, found {}".format(path, len(words)))
--- a/contrib/release-signing/sign-release-manifest.py
+++ b/contrib/release-signing/sign-release-manifest.py
@@ -52,13 +52,27 @@ MANIFEST_TAG = "Reddcoin/ReleaseManifest"
 
 # BIP32 path the release key is derived at.
 #
+#   m/1018'/4'/0'
+#     │      │   └── index, so a rotated key is 1' from the same backed-up seed
+#     │      └────── SLIP-0044 coin type 4, Reddcoin (chainparams nExtCoinType)
+#     └───────────── purpose
+#
 # Deliberately not a wallet path. Reddcoin wallets derive under m/44'/4'/..., and
 # a release key that could collide with a spending key is a bad idea however
 # unlikely the collision. This path exists for signing releases and nothing else.
 #
+# On the purpose value: 1018 was checked against the BIP43 registry (44, 45, 47,
+# 48, 49, 84, 86, 87) and against known unregistered users, and had no claimant.
+# Its neighbour 1017 does: lnd derives every Lightning key under
+# m/1017'/coinType'/keyFamily'/0/index, so that prefix would have described this
+# key as a Lightning multisig branch. Nothing would have collided, since this
+# seed derives nothing else, but a path is read by people and should not say
+# something untrue. Absence of a claimant is not a reservation; the REP is what
+# records this one.
+#
 # Fixed once a key has been generated: the mnemonic alone does not identify the
 # key without it, so changing the path is changing the key.
-DERIVATION_PATH = "m/1017'/4'/0'"
+DERIVATION_PATH = "m/1018'/4'/0'"
 
 
 def tagged_hash(tag, data):
--- a/contrib/release-signing/sign-release-manifest.py
+++ b/contrib/release-signing/sign-release-manifest.py
@@ -120,6 +120,18 @@ def bip39_seed(mnemonic, passphrase=""):
 
 
 def bip32_master(seed):
+    """BIP32 master key and chain code from a seed.
+
+    "Bitcoin seed" is the HMAC key BIP32 fixes for this step, in every
+    implementation and for every coin. It is not a name to localise: a coin
+    identifies itself in the derivation path, and changing this constant would
+    derive a different key from the same words while looking perfectly correct.
+    Keeping it standard is also what lets any BIP39 tool recover the release key
+    from the mnemonic, rather than this script being the only thing that can.
+
+    Matches CExtKey::SetSeed() in src/key.cpp, which spells the same constant as
+    a char array.
+    """
     i = hmac.new(b"Bitcoin seed", seed, hashlib.sha512).digest()
     return i[:32], i[32:]
 
--- a/contrib/release-signing/sign-release-manifest.py
+++ b/contrib/release-signing/sign-release-manifest.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2026 The Reddcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Sign and verify a release manifest with the offline release key.
+
+The client verifies releases with a BIP340 Schnorr signature over SHA256SUMS,
+checked against a public key compiled into the binary. This produces that
+signature. It is meant to run on the air-gapped machine that holds the private
+key, so it depends on nothing outside the Python standard library and this
+repository.
+
+The signing key never leaves that machine. Only SHA256SUMS goes in and only
+SHA256SUMS.sig comes out.
+
+Subcommands:
+
+    selftest   run the official BIP340 test vectors
+    pubkey     print the x-only public key for a mnemonic
+    sign       sign a manifest
+    verify     check a signature against a manifest and a public key
+
+Run selftest first on any machine you have not used before. It proves the
+implementation on that machine reproduces the BIP340 vectors, which is worth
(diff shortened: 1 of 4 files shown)


reddink:
Pushed `67c5e189ba`: the tool now validates the mnemonic before deriving.

As first written it derived a key from whatever words it was handed, with no
BIP39 wordlist or checksum check, deliberately, on the reasoning that the
printed public key is what tells you it is the right key. That guard is weaker
than it looks. It relies on a person comparing 64 hex characters by eye, and it
fails exactly where it matters most: recovering onto a rebuilt air-gapped
machine, where one mistyped word silently derives a perfectly usable key for a
wallet that has never signed anything.

The check is now mechanical:

```
$ ... pubkey --mnemonic valid.txt
2e1c1bc61241bb07be99d311605ca72405f117395915cec32b6d095788b00581

$ ... pubkey --mnemonic one-word-changed.txt      # exit 1
mnemonic checksum is wrong. One or more words are mistyped or out of order;
this is not the phrase you think it is

$ ... pubkey --mnemonic typo.txt                  # exit 1
not BIP39 words: abuot
```

The wordlist is parsed out of `src/util/lang/bip39_english.h`, the header the
client already ships, rather than copied into the script. A second copy could
drift from the one the wallet derives with, and the divergence would only ever
surface as a key that does not match.

Checked against the official BIP39 vectors: all five accepted at 12, 18 and 24
words, seed vector 1 reproduces the published 512-bit value exactly, and a
single-word corruption of each is rejected 5/5.

The public key check stays in the release process and is still the one that
cannot be skipped. A checksum proves a phrase is well formed, not that it is
the right phrase.

reddink:
Pushed three commits. The PR now matches REP-1018, which was merged into the reps repository as `docs/rep-1018.md`.

**`ac03997244` - the derivation path was wrong.** It was `m/1017'/4'/0'`, and `1017'` is lnd's: `BIP0043Purpose = 1017` in `keychain/derivation.go`, hierarchy `m/1017'/coinType'/keyFamily'/0/index`, `KeyFamilyMultiSig = 0`. So the path described the release key as a Lightning multisig branch for coin type 4. No key would ever have collided, since the release seed derives nothing else, but a derivation path is read by people and that one said something untrue. It is now `m/1018'/4'/0'`, checked against the BIP-allocated purposes, the SLIP-reserved 10001-19999 range, and known unregistered users. The number matches the REP that specifies it, which is what BIP43 asks for.

This was free to change only because no production key exists yet. One step later it would have cost a client release.

**`b5f0932c59`** documents why the BIP32 HMAC key stays the literal `"Bitcoin seed"`. It is fixed by BIP32 for every implementation and every coin; renaming it would derive a different key from the same words and put the key beyond the reach of every standard recovery tool.

**`a2c7faacb5` - the gap that mattered.** The signature is produced by Python on an air-gapped machine and checked by C++ in the client, and nothing forced the two to agree. A divergence would not have surfaced until every release failed to verify for every user.

`src/test/release_manifest_tests.cpp` asserts REP-1018 section 5.2's published vectors against the C++ side: `TaggedHash("Reddcoin/ReleaseManifest")` over the manifest, and the reference signature checked through the same `XOnlyPubKey::VerifySchnorr` call a client will make. The negative cases carry most of the value. A modified manifest, an appended line, a modified signature and a wrong key all fail, and so does a tag one character short or one character long, which is what shows the domain separation exists rather than being described in a comment.

Two traps worth naming, since both yield a test that passes while proving nothing. `uint256::ToString()` reverses byte order, so comparing it against the tool's hex tests a byte-swapped value. And `Span`'s container constructors are lifetime-bound, so building one from a `ParseHex` temporary risks a dangling diagnostic that is fatal under `-Werror`; every input in the test is a named local.

One thing deliberately not tested: `VerifySchnorr` asserts `sigbytes.size() == 64`, so a malformed `.sig` aborts the process rather than returning false. The length check is the caller's job, it belongs in the client code REP-1018 section 3.6 step 2 requires, and exercising it here would kill the test binary. There is a comment saying so.

Verified before pushing: `selftest` passes the 15 BIP340 vectors; the tool signs and verifies the real published 4.22.9.4 `SHA256SUMS`, with five tamper cases each exiting 1; mnemonic validation rejects a bad checksum, a non-BIP39 word and a wrong length, and neutering the check makes the mistyped phrase derive a different key, which is what proves the check is load bearing.

CI is the first thing to compile the new test. I have not built it locally.

reddink:
Built and ran on both lines. `release_manifest_tests` passes 9/9 on each.

```
develop line   [contrib/release-manifest-signing]
  release_manifest_tests     9 cases, exit=0 PASS
  release_artifacts_tests    9 cases, exit=0 PASS
  update_check_tests        12 cases, exit=0 PASS

4.22.9 line    [contrib/release-manifest-signing-4.22.9]
  release_manifest_tests     9 cases, exit=0 PASS
  release_artifacts_tests    9 cases, exit=0 PASS
```

It compiles clean on both, so the earlier caveat about CI being the first thing to build it no longer applies.

**The test was also checked for teeth**, since a cross-implementation guard that passes regardless of the truth is worse than no guard. Flipping the last character of the expected tagged hash and rebuilding produces:

```
error: in "release_manifest_tests/cpp_and_the_signing_tool_agree_on_the_message":
  check RawHex(ManifestHash(TAG, MANIFEST)) == EXPECTED_TAGGED_HASH has failed
  [f0b197e4...dcaf480 != f0b197e4...dcaf481]
exit=201
```

The value on the left is what the C++ computed on its own, and it is exactly what the Python tool produces. That is the agreement this test exists to establish, measured rather than argued. Reverted and rebuilt, back to green.

On the 4.22.9 side specifically, the tool was run from that checkout and resolves the BIP39 wordlist through `src/wallet/bip39_english.h`, deriving the same public key from the same mnemonic as it does on develop. Both wordlists agree, and the file stays byte identical across the two lines.

PR #524, opened 2026-09-04
contrib: sign the release manifest for the client

reddink:
Backport of #523 to the maintenance line. Phase 2b of RED-98 (RED-113).

This line is where releases are actually cut from, so it matters more here than on develop.

The format is specified in **REP-1018** (`docs/rep-1018.md` in the reps repository) and the purpose value it derives under is recorded in **REP-0005**, the derivation path registry.

## What comes across

- `contrib/release-signing/sign-release-manifest.py`, the offline tool: `selftest`, `pubkey`, `sign`, `verify`. Stdlib and this repository only, so it runs on an air-gapped box with nothing installed.
- `src/test/release_manifest_tests.cpp`, which holds the C++ side to the same answer as the Python signer. Nothing else forces the two to agree, and a divergence would not surface until every release failed to verify for every user.
- The publishing, signing and verification sections of `doc/release-process.md`.

## Why the doc change is bigger than the tool

Deliberate, and the part most worth reviewing.

This line's release-process still says to build the published manifest with:

```bash
cat "$VERSION"/*/all.SHA256SUMS.asc > SHA256SUMS.asc
```

That signs a file which is not the one served. `all.SHA256SUMS` carries the codesignature and debug entries; the published `SHA256SUMS` does not. The difference varies per release, so it can appear to work on one release and produce a **BAD signature** on the next.

Grafting a new client-signing step onto a procedure that already produces a bad `.asc` would have left this branch documenting two signatures over different bytes, on the line that ships releases. So the sections come across whole and this file now matches develop exactly.

## Line portability

The tool parses the client's own BIP39 header rather than carrying a second copy that could drift. That header is at `src/wallet/bip39_english.h` here and `src/util/lang/bip39_english.h` on develop, so it tries both. The file is therefore **byte identical on both lines** rather than forked, which matters for something meant to be copied to an air-gapped machine.

## Verified on this branch

- BIP340 self-test passes its 15 official vectors
- the tool derives the **same public key from the same mnemonic** as it does on develop, which is what shows both wordlists agree and the path resolution works here
- mnemonic validation rejects a bad checksum, a non-BIP39 word and a wrong length, each exit 1
- prerequisites confirmed present on this line: `TaggedHash` in `src/hash.cpp`, `XOnlyPubKey::VerifySchnorr`, `ECCVerifyHandle` in the test fixture, and `--enable-module-schnorrsig` already in `configure.ac`

The unit test has not been compiled on this branch. CI is the first thing to build it.

## Not in this PR

The production key does not exist yet; `doc/release-process.md` carries a placeholder for its public key. Client-side verification is phase 4 and must not be written until a real `.sig` is published.

YouTrack: https://reddink.youtrack.cloud/issue/RED-113

### DIFF
--- a/contrib/release-signing/sign-release-manifest.py
+++ b/contrib/release-signing/sign-release-manifest.py
@@ -1,0 +1,319 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2026 The Reddcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Sign and verify a release manifest with the offline release key.
+
+The client verifies releases with a BIP340 Schnorr signature over SHA256SUMS,
+checked against a public key compiled into the binary. This produces that
+signature. It is meant to run on the air-gapped machine that holds the private
+key, so it depends on nothing outside the Python standard library and this
+repository.
+
+The signing key never leaves that machine. Only SHA256SUMS goes in and only
+SHA256SUMS.sig comes out.
+
+Subcommands:
+
+    selftest   run the official BIP340 test vectors
+    pubkey     print the x-only public key for a mnemonic
+    sign       sign a manifest
+    verify     check a signature against a manifest and a public key
+
+Run selftest first on any machine you have not used before. It proves the
+implementation on that machine reproduces the BIP340 vectors, which is worth
+knowing before it signs something users will trust.
+"""
+
+import argparse
+import hashlib
+import hmac
+import os
+import re
+import sys
+import unicodedata
+
+# The BIP340 implementation, the curve order and the official test vectors all
+# already live in the test framework. Reusing them beats a second copy of
+# signature code that nothing else exercises.
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'test', 'functional'))
+from test_framework.key import (  # noqa: E402
+    SECP256K1_ORDER,
+    compute_xonly_pubkey,
+    sign_schnorr,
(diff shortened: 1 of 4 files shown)


reddink:
Built and ran on both lines. `release_manifest_tests` passes 9/9 on each.

```
develop line   [contrib/release-manifest-signing]
  release_manifest_tests     9 cases, exit=0 PASS
  release_artifacts_tests    9 cases, exit=0 PASS
  update_check_tests        12 cases, exit=0 PASS

4.22.9 line    [contrib/release-manifest-signing-4.22.9]
  release_manifest_tests     9 cases, exit=0 PASS
  release_artifacts_tests    9 cases, exit=0 PASS
```

It compiles clean on both, so the earlier caveat about CI being the first thing to build it no longer applies.

**The test was also checked for teeth**, since a cross-implementation guard that passes regardless of the truth is worse than no guard. Flipping the last character of the expected tagged hash and rebuilding produces:

```
error: in "release_manifest_tests/cpp_and_the_signing_tool_agree_on_the_message":
  check RawHex(ManifestHash(TAG, MANIFEST)) == EXPECTED_TAGGED_HASH has failed
  [f0b197e4...dcaf480 != f0b197e4...dcaf481]
exit=201
```

The value on the left is what the C++ computed on its own, and it is exactly what the Python tool produces. That is the agreement this test exists to establish, measured rather than argued. Reverted and rebuilt, back to green.

On the 4.22.9 side specifically, the tool was run from that checkout and resolves the BIP39 wordlist through `src/wallet/bip39_english.h`, deriving the same public key from the same mnemonic as it does on develop. Both wordlists agree, and the file stays byte identical across the two lines.

PR #536, opened 2026-09-05
node: verify a release manifest against the pinned key

reddink:
Phase 4a of RED-98, first slice of RED-119. Adds the verification a downloaded artifact has to pass before anyone is offered it. Phase 3 can fetch a file; until this exists, that file is untrusted bytes from the network.

Implements the checkable half of REP-1018 section 3.6: verify the signature over the manifest, then resolve one artifact name to the digest the manifest gives for it. Downloading the artifact and comparing its hash is 4b.

## The trust anchor

```cpp
const std::string node::RELEASE_PUBKEY{
    "23e6b696f2b69cd753de0d8fe3875e989085fbb6f2dc09026ba8ba1df33585db"};
```

Compiled in, which is the whole design: no keyserver, no handshake, no network lookup. A hostile network can deny an upgrade but cannot substitute one. Its private half exists only on the air-gapped machine, at `m/1018'/4'/0'`.

## Three things that are load bearing, not defensive

**A signature that does not decode to exactly 64 bytes is rejected before verification is attempted.** `XOnlyPubKey::VerifySchnorr` asserts on any other length, so reaching it with a malformed `.sig` would **abort the process** rather than report a bad release.

**The filename match is on the whole field, and requires exactly one line.** The published manifest lists signed and unsigned variants side by side:

```
...  reddcoin-4.22.9.4-win64-setup-signed.exe
...  reddcoin-4.22.9.4-win64-setup-unsigned.exe
```

A substring match for `signed.exe` matches both. A client matching loosely would hand a Windows user the **unsigned installer while reporting a verified download**. A duplicated entry fails rather than picking one, because which digest is right is exactly what a duplicate makes unknowable.

**The manifest is hashed over the bytes as received.** Normalising line endings or whitespace on the way in would verify something the server does not serve.

## The test that matters

`the_real_published_release_verifies` embeds the **actual published 4.22.9.4 manifest and its signature**, byte for byte, and checks they verify against the pinned key. It passes, so the key, the tag, the message construction and the parser agree with what the release process really produced rather than merely with each other.

It also confirms the digest the manifest gives for this host's artifact is the one that **downloading that artifact actually produced** in #531, and that a single flipped byte or an appended line breaks verification.

Offline on purpose: a regression test, not a liveness check.

## Checked for teeth

Flipping the last character of the pinned key fails that test and two others:

```
error: the real release failed to verify: Release public key is not a valid point
*** 3 failures detected
```

So the anchor is doing work rather than being decoration.

## Verified

10 new tests, plus 9 + 9 + 27 existing, all passing on both lines, builds clean including `reddcoin-qt`.

## Not in this PR

Downloading the artifact and comparing its hash, and the staging lifecycle. That is 4b, and it is a small addition now that this and phase 3b both exist.

YouTrack: https://reddink.youtrack.cloud/issue/RED-119

---

**Backport.** All three new files are byte identical to develop's, which is what keeps 4b a copy across rather than a translation. Verified on this branch: build clean including `reddcoin-qt`, 10 + 9 + 9 + 27 tests pass.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -186,6 +186,7 @@ BITCOIN_CORE_H = \
   node/blockstorage.h \
   node/ca_store.h \
   node/release_artifacts.h \
+  node/release_verify.h \
   node/update_check.h \
   node/coin.h \
   node/coinstats.h \
@@ -365,6 +366,7 @@ libbitcoin_server_a_SOURCES = \
   node/blockstorage.cpp \
   node/ca_store.cpp \
   node/release_artifacts.cpp \
+  node/release_verify.cpp \
   node/update_check.cpp \
   node/coin.cpp \
   node/coinstats.cpp \
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -146,6 +146,7 @@ BITCOIN_TESTS =\
   test/txvalidationcache_tests.cpp \
   test/release_artifacts_tests.cpp \
   test/release_manifest_tests.cpp \
+  test/release_verify_tests.cpp \
   test/update_check_tests.cpp \
   test/uint256_tests.cpp \
   test/util_tests.cpp \
--- a/src/node/release_verify.cpp
+++ b/src/node/release_verify.cpp
@@ -1,0 +1,181 @@
+// Copyright (c) 2014-2026 The Reddcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <node/release_verify.h>
+
+#include <crypto/sha256.h>
+#include <hash.h>
+#include <pubkey.h>
+#include <span.h>
+#include <streams.h>
+#include <util/strencodings.h>
+#include <util/string.h>
+
+#include <array>
+#include <string>
+#include <vector>
+
+namespace {
+//! Domain separation for a release signature, per REP-1018 section 3.4.
+//!
+//! The release key sits on the same curve as transaction keys, so a signature
+//! over a bare SHA256 could be meaningful in another context. The tag is part
+//! of the format: changing a byte of it invalidates every signature any
+//! released client will accept.
+const CHashWriter HASHER_RELEASE_MANIFEST = TaggedHash("Reddcoin/ReleaseManifest");
+
+//! The message a release signature commits to.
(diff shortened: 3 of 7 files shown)


reddink:
**Phase 4b pushed to both PRs**, completing phase 4. A downloaded file is now either proven to be the artifact the release key signed, or it is deleted.

`StageVerifiedRelease` performs the whole of REP-1018 section 3.6 end to end: fetch the manifest and signature, verify the signature against the compiled-in key, resolve the artifact name to its digest, download via the phase 3b streaming path, and hash what arrived.

## Three lifecycle decisions, settled here rather than in phase 5

**One directory per release, other versions pruned on success.** Disk use is bounded by a single release rather than growing with every check. Pruning happens **after** a success, never before the download, so a failed fetch cannot destroy a good artifact that is already staged.

**Reuse is by hash, not by name.** An artifact already staged and already correct is not downloaded again, so repeating the check costs a manifest rather than 29 MB. A file present under the right name with the wrong contents is deleted and refetched, because the only thing knowable about it is that it is not the artifact.

**A hash mismatch deletes the file.** Leaving it would put an unverified artifact in the staging directory looking exactly like a verified one, which is the failure this phase exists to prevent.

## Verified against the live server

A temporary network test, removed before commit:

```
staged .../updates/reddcoin-core-4.22.9.4/reddcoin-4.22.9.4-x86_64-linux-gnu.tar.gz
       (30603976 bytes) in 10632 ms, 1946 progress ticks
second run: 2662 ms, 0 ticks          <- reused, no re-download
corrupted file was refetched
stale version directory pruned
unlisted artifact: The release manifest does not list no-such-artifact.tar.gz
```

The second run is the one worth looking at: zero progress ticks means the 29 MB was not fetched again, only the manifest.

## Tests

4 added, 14 in the suite. The download needs a network, so what is tested offline is the part that can go wrong quietly: that verification precedes staging, that a file is recognised by its hash rather than its name, and that **two files of identical length and different content do not hash alike**, since a size check is the tempting shortcut and would pass both.

All four suites pass on both lines: 14 + 9 + 9 + 27, clean builds including `reddcoin-qt`.

## Phase 4 is complete

What remains is phase 5: presenting this to the user, launching the installer, and the progress dialog. It will find the staging lifecycle already decided rather than having to invent it alongside a UI.

PR #539, opened 2026-09-05
qt: offer to download and verify the update from the dialog

reddink:
Phase 5b of RED-98, second slice of RED-120. The update dialog named the build for this machine and linked to it; now it can fetch that build and prove it is the one the release key signed, without the user leaving the client.

**Nothing is installed.** The dialog ends at "here is a verified file", which is Tier 2 as agreed.

## The button appears only where an artifact was named

An up-to-date node, and a release with no artifact for this host, both keep exactly the dialog they had. The case that cannot be handled honestly is not offered, rather than being offered and failing.

## Both phase 3b assumptions were exercised, and both held

They were decided when nothing called the download API, so this is the first time they have been tested.

**The staging call blocks** for as long as a 29 MB transfer takes, so it runs on a worker thread and reaches the dialog through queued signals.

**Progress fires around 180 times a second.** That is reasonable for a callback and far too high to drive a repaint, so it is thinned to whole percentage points **in the worker**, bounding the queued signal traffic at the source rather than leaving every consumer to remember. Where the server declares no length there is no honest percentage, so the bar goes indeterminate rather than inventing a position.

## Cancelling

Sets a flag the worker polls between reads. The transfer stops at the next one rather than being torn down from under the thread performing it, which is the failure `Await()`'s shutdown dance exists to avoid and which cancel would otherwise have reintroduced.

The destructor cancels **before** waiting. `quit()` alone cannot reach a worker sitting inside a blocking call, so closing the dialog mid-download would otherwise block on a transfer nobody wants any more.

## Tests

Four Qt tests, none touching the network. They cover the wiring, which is where this goes wrong quietly: that the controls appear only when there is something to fetch, and that a dialog carrying a worker thread can be constructed and destroyed repeatedly without incident.

## Not visually confirmed

Stated plainly because it is the honest limit here. The Qt built by `depends` ships only the `xcb` platform plugin, so there is no headless mode to drive the dialog in. The automated tests cover the wiring; what they cannot cover is whether it looks right. Worth a look in a running client before merge.

YouTrack: https://reddink.youtrack.cloud/issue/RED-120

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -85,6 +85,7 @@ QT_MOC_CPP = \
   qt/moc_transactiontablemodel.cpp \
   qt/moc_transactionview.cpp \
   qt/moc_updatecheckworker.cpp \
+  qt/moc_updatedownloadworker.cpp \
   qt/moc_utilitydialog.cpp \
   qt/moc_walletcontroller.cpp \
   qt/moc_walletframe.cpp \
@@ -166,6 +167,7 @@ BITCOIN_QT_H = \
   qt/transactiontablemodel.h \
   qt/transactionview.h \
   qt/updatecheckworker.h \
+  qt/updatedownloadworker.h \
   qt/utilitydialog.h \
   qt/walletcontroller.h \
   qt/walletframe.h \
@@ -263,6 +265,7 @@ BITCOIN_QT_BASE_CPP = \
   qt/splashscreen.cpp \
   qt/trafficgraphwidget.cpp \
   qt/updatecheckworker.cpp \
+  qt/updatedownloadworker.cpp \
   qt/utilitydialog.cpp
 
 BITCOIN_QT_WINDOWS_CPP = qt/winshutdownmonitor.cpp
--- a/src/Makefile.qttest.include
+++ b/src/Makefile.qttest.include
@@ -8,7 +8,8 @@ TESTS += qt/test/test_reddcoin-qt
 TEST_QT_MOC_CPP = \
   qt/test/moc_apptests.cpp \
   qt/test/moc_rpcnestedtests.cpp \
-  qt/test/moc_uritests.cpp
+  qt/test/moc_uritests.cpp \
+  qt/test/moc_updatedialogtests.cpp
 
 if ENABLE_WALLET
 TEST_QT_MOC_CPP += \
@@ -21,6 +22,7 @@ TEST_QT_H = \
   qt/test/apptests.h \
   qt/test/rpcnestedtests.h \
   qt/test/uritests.h \
+  qt/test/updatedialogtests.h \
   qt/test/util.h \
   qt/test/wallettests.h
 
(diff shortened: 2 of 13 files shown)


reddink:
**Pushed a second commit, and it is a bug fix that this PR needed.**

While backporting 5b I found that the download button would never have appeared on develop at all, for a reason that predates this PR.

`UpdateCheckWorker` named seven fields to copy out of the node's result. `showUpdateInfo` had grown to read ten. So `guiartifact`, `guiartifactlink` and `platform` arrived empty, and the dialog took its "no artifact is published for this host" branch on **every** update.

Two consequences:

- **The named-file notice phase 1 added has never worked through the real path on develop.** The dialog has been showing the generic directory link since #517.
- This PR's download button sits in the branch that was never reached, so it would have shipped invisible.

## Why nothing caught it

Every test of the dialog, including the three I added here, hands it a `QVariantMap` built in the test. That proves the dialog reads a map correctly and proves nothing about whether the worker ever puts those keys in one. Both sides were tested; the seam between them was not.

## The fix

Copy every field rather than naming the wanted ones. An allow-list drifts the moment the node reports something new, and it drifts **silently**: the reader goes on asking for a key nobody puts in, gets an empty string, and takes whichever branch empty means. Copying everything cannot drift.

The new test asserts the seam rather than either side of it. Reverting the mapping fails it with:

```
FAIL!  : everyFieldTheDialogReadsSurvivesTheWorker() 'info.contains(...)' returned FALSE.
         (the worker dropped 'hosttriplet')
```

while the other three dialog tests still pass, which is exactly the blind spot it exists to cover.

Worth reviewing as its own commit. It is a defect in shipped behaviour, not part of 5b, and it should probably be understood separately even though it lands here.

⚠ The 4.22.9 line is **not** affected: it reads the fields straight out of the UniValue in the dialog, so it never had this seam. Its problem is the opposite one, that the read is synchronous on the GUI thread, tracked as RED-121.

reddink:
**Phase 5c pushed here too**, so this branch now carries three separable commits. Worth reviewing as three things rather than one:

| Commit | |
| --- | --- |
| `53395a75fc` | 5b, the download dialog |
| `2807ef4826` | the RED-123 fix: the worker dropped fields, which had disabled phase 1's named-file notice and would have made 5b's button invisible |
| `8ef97ea7dd` | 5c, the platform hand-off |

## 5c

Once a download is verified, the button becomes the platform's next step: **Run installer** on Windows, **Open disk image** on macOS, **Show in folder** on Linux.

Linux stops at revealing the file deliberately. The install shape is not knowable from inside the client, so opening the archive would hand it to an archive manager, which is a guess dressed up as help. That is B5.

Nothing is installed and nothing is executed on the user's behalf. The tooltip says the client will not close itself, because on Windows and macOS the next thing that happens is an installer wanting to replace a running application. A hand-off that fails says so rather than leaving a button that silently does nothing.

The test asserts the wording matches the platform, and specifically that the Linux label does not promise to install anything.

## ⚠ Stacked

This branches from `rpc/download-update` (#537). Merge **#537 first**, then this. That will shrink this PR's diff to its own three commits.

**This completes RED-98.** Phases 0 through 5 are all implemented on both lines.

PR #548, opened 2026-09-07
validation: verify proof of work in CheckBlockHeader

reddink:
`CheckBlockHeader` on this line is a stub:

```cpp
static bool CheckBlockHeader(const CBlockHeader& block, BlockValidationState& state,
                             const Consensus::Params& consensusParams, bool fCheckPOW = true)
{
    return true;
}
```

`CheckProofOfWork` is therefore called nowhere in validation, and `CBlockHeader::GetPoWHash`, the scrypt hash, is defined but never called from anywhere. A block's claimed `nBits` is never tested against any actual work.

`80b9c562c6`, the PoSV migration, removed the check rather than adapting it. The upstream line it deleted was wrong for Reddcoin, because it hashed `GetHash()` where our proof of work is scrypt, but deleting it left nothing in its place. develop has since restored it correctly, and this is that fix ported across.

> Stacked on #547, which it needs for the chain fixture. Review that one first.

## What this actually reaches, which is narrower than it sounds

`CheckBlockHeader` takes a `CBlockHeader`, so `block.IsProofOfWork()` binds to the header predicate, which is `nVersion <= POW_BLOCK_VERSION`, and not to the coinstake predicate `CBlock` shadows it with. Every block the miner produces carries a versionbits `nVersion` of `0x20000000`, so the check does not apply to them.

What it covers is the historical proof-of-work era: heights at or below `nLastPowHeight`, which is 260799 on mainnet, where blocks carry version 1 or 2 and are dated after `CHECK_POW_FROM_NTIME`. That is the range that was unprotected, and it is the same reach the check has on develop, so this does not make the two lines diverge.

Exposure before this change was limited rather than open, and worth stating plainly so the change is not over-rated. Proof-of-stake blocks were never affected, since `CheckProofOfStake` tests the kernel against a target derived from the same `nBits` that chainwork is computed from. The whole proof-of-work era sits below the last mainnet checkpoint, at height 5728519, and `ContextualCheckBlockHeader` rejects forks below it. This restores defence in depth that had been lost, and removes a dependency on the checkpoint list that was never meant to be load bearing.

## Commits

1. **`validation: verify proof of work in CheckBlockHeader`** restores the check, hashing `GetPoWHash()`, skipping proof-of-stake headers, and exempting blocks at or before `CHECK_POW_FROM_NTIME`. Adds that constant, matching develop.

2. **`mining: grind the scrypt hash rather than SHA256d`** fixes both nonce-grinding loops, in `rpc/mining.cpp` and the test fixture, which tested `CheckProofOfWork` against `block.GetHash()`. They were satisfying a target the consensus code does not check. Harmless only while `CheckBlockHeader` was a stub; now that it verifies work, grinding the wrong hash could produce blocks the node itself rejects.

   Grinding a different hash selects a different nonce, so the fixture's chain has different block hashes and the tip hash recorded in #547 is updated here. Confirmed identical across separate runs and separate suites. The chain still reaches height 60, so nothing is rejected by the restored check.

3. **`test: cover the proof of work check in CheckBlockHeader`** adds the regression test the original check never had, which is how it came to be deleted without anything failing.

## The test

`CheckBlockHeader` is static, but `CheckBlock` is not and calls it before any check that inspects transactions, so a block carrying only a header reaches it. Four cases, covering the check and each condition that switches it off:

- a proof-of-work header missing its target is rejected as `high-hash`
- one meeting its target is not, so the first case is not merely watching the check reject everything put in front of it
- a proof-of-stake header is exempt, decided by the header's version rather than the coinstake test, which pins down why the check does not apply to blocks produced today
- a header at `CHECK_POW_FROM_NTIME` is exempt, matching the historical chain

Each case asserts its own premise, that the header really does miss or meet `CheckProofOfWork`, so none can quietly stop testing anything if the hash or the target changes. They use `BasicTestingSetup` and build no chain, so they are independent of the chain fixture.

## Verification

Four new cases pass. Full Qt suite green, exit 0. `blockfilter_index_tests`, `txvalidationcache_tests`, `coinstatsindex_tests` and `txindex_tests` all still pass.

## Pre-existing failures, not from this change

Three suites fail on this line for reasons this PR does not touch, all of them unadapted upstream tests that develop has already adapted, and all belonging with RED-125:

- `pow_tests` has 5 failing cases that assert Bitcoin's `CalculateNextWorkRequired` values. Reddcoin uses Kimoto Gravity Well and that function is a stub returning 0. develop expects `0U` and adds `get_next_work_kgw`. The changes here are pure additions to that file, 99 insertions and no deletions.
- `validation_tests/block_subsidy_test` asserts Bitcoin's 50 COIN initial subsidy against Reddcoin's schedule. develop adapted it to the real halvings at 140,000 and 190,000.
- `validation_block_tests` fails with `BLOCK_RESULT_UNSET` and an empty reject reason, so it is not the check added here, which can only fail with `high-hash`. It uses `MinerTestingSetup` rather than the chain fixture, and develop carries a 59-line adaptation of the file that this line lacks.

https://reddink.youtrack.cloud/issue/RED-124

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -82,7 +82,9 @@ int64_t UpdateTime(CBlockHeader* pblock, const Consensus::Params& consensusParam
 void RegenerateCommitments(CBlock& block, ChainstateManager& chainman)
 {
     CMutableTransaction tx{*block.vtx.at(0)};
-    tx.vout.erase(tx.vout.begin() + GetWitnessCommitmentIndex(block));
+    if (GetWitnessCommitmentIndex(block) != NO_WITNESS_COMMITMENT) {
+        tx.vout.erase(tx.vout.begin() + GetWitnessCommitmentIndex(block));
+    }
     block.vtx.at(0) = MakeTransactionRef(tx);
 
     CBlockIndex* prev_block = WITH_LOCK(::cs_main, return chainman.m_blockman.LookupBlockIndex(block.hashPrevBlock));
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -134,7 +134,10 @@ void BumpFee(TransactionView& view, const uint256& txid, bool expectDisabled, st
 //     QT_QPA_PLATFORM=cocoa   src/qt/test/test_bitcoin-qt  # macOS
 void TestGUI(interfaces::Node& node)
 {
-    // Set up wallet and chain with 105 blocks (5 mature blocks for spending).
+    // Set up wallet and chain. TestChain100Setup mines nCoinbaseMaturity (60)
+    // blocks here, at which height no coinbase has matured yet, so five more
+    // are needed before anything is spendable. Heights 61-65 are still at or
+    // below nLastPowHeight = 89, so these are valid PoW blocks.
     TestChain100Setup test;
     for (int i = 0; i < 5; ++i) {
         test.CreateAndProcessBlock({}, GetScriptForRawPubKey(test.coinbaseKey.GetPubKey()));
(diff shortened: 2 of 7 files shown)


reddink:
Correction to this PR's description, for the record.

The description lists a fourth change, `pow: reject negative and overflowing compact targets" in `src/pow.cpp`. **This PR did not contain it.** The commit was written after this branch was pushed and never re-pushed here, so it merged in #549 instead, which was based on the local tip of this branch.

The fix is in `v4.22.9-regtest` and nothing is missing. But anyone reading this PR to see how `CheckProofOfWork` was hardened should look at #549, not here, and `git blame` on `src/pow.cpp` will point there too.

PR #549, opened 2026-09-07
test: adapt upstream expectations to Reddcoin's rules

reddink:
Several test files on this line still assert Bitcoin's numbers against Reddcoin's consensus rules, so they fail on arithmetic that was never going to hold. develop has already adapted them; this ports those adaptations across.

> Stacked on #548, which is stacked on #547. Review those first.

Scope is deliberately limited to ports. Where develop's version is built on its staking fixture, or hardcodes values for the chain that fixture produces, it is not taken: `TestChain100Setup` on this line mines `nCoinbaseMaturity` blocks, which is 60 on regtest, and reaching 100 needs proof-of-stake blocks for heights 90 to 100. That decision is recorded in RED-125.

## Ported

**`pow_tests`** expected Bitcoin's retarget results from `CalculateNextWorkRequired`. Reddcoin uses Kimoto Gravity Well and that function is a stub returning 0, so the four cases built on it could not pass. Replaced with develop's `calculate_next_work_stub` and `get_next_work_kgw`, which test the retarget actually in use.

**`validation_tests`** expected Bitcoin's 50 COIN initial subsidy and halving schedule. Reddcoin pays 10,000 RDD at genesis, 545,000,000 for the premine blocks, then a bonus schedule before the halvings at 140,000 and 190,000. Replaced `block_subsidy_test` and `subsidy_limit_test` with develop's, which encode the real schedule.

**`interfaces_tests`** asserted the tip was at height 100 and then indexed the chain at 99 and 100. On a 60-block chain the assertion was wrong and the indexing walked off the end, which is what turned that failure into a segfault. Takes the height from the chain now, as develop does.

**`wallet_tests`** compared against an error string still naming `bitcoind` where the code emits `reddcoind`.

## Result

`pow_tests` passes in full, 17 cases, including the four added in #548 and `CheckProofOfWork_test_negative_target`, which that PR's `pow.cpp` fix is what repairs.

`validation_tests` goes from 3168 failures to 6. `interfaces_tests` from 2 to 1.

`blockfilter_index_tests`, `txvalidationcache_tests`, `coinstatsindex_tests` and `txindex_tests` still pass.

## Left failing, deliberately

**`validation_tests/test_assumeutxo`**, 3 checks. The entries in this line's chainparams are for heights 110 and 200, which the fixture cannot reach, and `src/chainparams.cpp` already carries a `TODO: Recompute these hashes after RegenerateCommitments fix`. develop's version of this test hardcodes hashes for its own chain, so porting it would swap one set of wrong constants for another.

**`validation_tests/signet_parse_tests`**, 3 checks. develop deleted this test; it is kept here. It forces `-signetchallenge` to `51` and still does not get `OP_TRUE` back from `CreateChainParams`, which suggests the argument is not being honoured. That is a real question, and deleting a red test without understanding it would hide the answer rather than produce one.

**`interfaces_tests/hasBlocks`**, segfault. It clears `BLOCK_HAVE_DATA` on `active[95]`, which is null on a 60-block chain. develop's copy is byte-identical and passes only because its chain reaches 100, so there is nothing to port: adapting it means choosing new sample heights, which is a rewrite rather than a port.

**`wallet_tests`** balance and coinbase-count expectations, and **`validation_chainstatemanager_tests`**. develop's versions of both call `stakeBlocks` throughout. The latter also needs the chain at height 110 for its assumeutxo entry, and proof of work is capped at `nLastPowHeight`, which is 89, so that case cannot pass without proof-of-stake blocks at all.

**`miner_tests`** is unchanged at 115 failures on a different fixture, and is not part of this.

https://reddink.youtrack.cloud/issue/RED-125

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -137,11 +137,18 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
 
 bool CheckProofOfWork(arith_uint256 hash, unsigned int nBits, const Consensus::Params& params)
 {
+    bool fNegative;
+    bool fOverflow;
     arith_uint256 bnTarget;
-    bnTarget.SetCompact(nBits);
 
-    // Check range
-    if (bnTarget <= 0 || bnTarget > UintToArith256(params.powLimit))
+    bnTarget.SetCompact(nBits, &fNegative, &fOverflow);
+
+    // Check range. arith_uint256 is unsigned, so bnTarget <= 0 only ever caught
+    // zero: a compact encoding carrying the sign bit was silently accepted as
+    // its own magnitude, and one whose exponent overflows was accepted as
+    // whatever SetCompact left behind. Both have to come from SetCompact's own
+    // flags, which is what upstream does and what develop already carries.
+    if (fNegative || bnTarget == 0 || fOverflow || bnTarget > UintToArith256(params.powLimit))
         return error("CheckProofOfWork() : nBits below minimum work");
 
     // Check proof of work matches claimed amount
--- a/src/test/interfaces_tests.cpp
+++ b/src/test/interfaces_tests.cpp
@@ -46,13 +46,17 @@ BOOST_AUTO_TEST_CASE(findBlock)
 
     bool cur_active{false}, next_active{false};
     uint256 next_hash;
-    BOOST_CHECK_EQUAL(active.Height(), 100);
-    BOOST_CHECK(chain->findBlock(active[99]->GetBlockHash(), FoundBlock().inActiveChain(cur_active).nextBlock(FoundBlock().inActiveChain(next_active).hash(next_hash))));
+    // TestChain100Setup mines nCoinbaseMaturity blocks on this line, which is
+    // 60 on regtest, so the tip is not at the height the fixture's name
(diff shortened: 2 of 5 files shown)


reddink:
Correction to this PR's description, for the record.

This PR also carried `62f6a93360` "pow: reject negative and overflowing compact targets", a change to `src/pow.cpp`, which the description does not mention and which was described in #548 instead.

That commit was written for #548 but never pushed to its branch. This branch was based on the local tip of that branch, which already had it, so it travelled here instead. #548 merged without it.

Nothing is missing from `v4.22.9-regtest`: the fix is in the merged tree and `pow_tests` passes, including `CheckProofOfWork_test_negative_target`, which is the case that fails without it. But a consensus-code change landed in a pull request whose description said test adaptations only, and reviewers had no reason to look for it there.

The change itself: `CheckProofOfWork` validated its decoded target with `bnTarget <= 0`, and `arith_uint256` is unsigned, so that only ever caught zero. A compact encoding carrying the sign bit was accepted as its own magnitude, and one whose exponent overflows was accepted as whatever `SetCompact` left behind. Both now come from `SetCompact`'s own flags, matching upstream and develop.

PR #558, opened 2026-09-10
qt: refresh only the transaction rows a new block can change

reddink:
On a wallet with a long staking history, reddcoin-qt froze for about ten seconds after every block. The transaction table was re-evaluating every row it holds, in every filter proxy attached to it, on the GUI thread, once per wallet per block.

Tracked as [RED-134](https://reddink.youtrack.cloud/issue/RED-134), part of the [RED-130](https://reddink.youtrack.cloud/issue/RED-130) epic on GUI stalls with large staking wallets. This is the release line first; the port to `develop` follows.

## The problem

`TransactionTableModel::updateConfirmations` runs on every tip change and emitted `dataChanged` over the whole table, twice. Each `TransactionFilterProxy` on the model (the history view and the overview list, both with dynamic sorting) answers a `dataChanged` by re-running `filterAcceptsRow` on every row in the range. Per row that meant a status refresh with three `cs_main` acquisitions, seven role lookups, a base58 decode of the address, and a hard `cs_wallet` lock for the label. Four full passes per wallet per block, on every loaded wallet, shown or not.

A staking wallet gains a transaction per stake, so it reaches sizes this code was never run at. Measured on a live mainnet client with the timer from the first commit, `--enable-debug` build:

| Wallet | Rows | `updateConfirmations` per block |
|---|---|---|
| Main_funds | 176,005 | 4,921 ms |
| default wallet | 59,093 | 1,601 ms |
| Bittrex_Funds_1 | 38,054 | 1,074 ms |
| all 16 loaded wallets | 273,152 | 7,611 ms |

A flat 27 to 28 us per row, and the GUI thread at 100% for 8 to 10 s after each block.

## The change

Only rows waiting on depth can change with a block without the wallet reporting it: unconfirmed, confirming, immature, and the two non-final states. `updateConfirmations` now scans the cached records in memory for those and emits per contiguous run. A confirmed row only gains confirmations, which `index()` already refreshes lazily when the row is next painted or hovered. A conflicted, abandoned or rejected row moves only through a transaction notification, so `updateWallet` now re-evaluates those rows itself when the notification arrives; previously it set a flag and left the whole-table sweep to notice.

A changed row emits its full column span rather than the status and address cells, so the amount brackets and the row colour repaint with the status icon. `needsUpdate` was never initialised and now is.

Two smaller changes cut the per-row cost of the passes that remain (the mapping build when a wallet loads, and filter changes in the history view):

- `labelForAddress` answers from a label index kept in `AddressTableModel`, maintained by the same address book notifications as the table, instead of decoding the address and taking `cs_wallet` on every call. The index is deliberately not narrowed by `pk_hash_only`: the sign-message dialog rebuilds the wallet's address table with pay-to-pubkey-hash entries only, and labels for other address types must survive that. A miss re-spells the address and looks again, so user input in another spelling still resolves.
- `filterAcceptsRow` builds the address, label and txid strings only when a search string is set.

The BENCH timer from the first commit stays, reporting live rows, total rows and range count per wallet per block under `-debug=bench`.

## Result

Same client, same wallets, staking on:

| | Before | After |
|---|---|---|
| Transaction-table refresh, all wallets, per block | 7,611 ms | about 3 ms |
| GUI thread pinned at 100% after a block | 8 to 10 s | none |
| GUI-thread CPU per block | about 8 s | 1.5 to 3 s |

The remaining 1.5 to 3 s is the stake-weight recomputation on the GUI thread, which is RED-131.

On a testnet wallet with 1,181,130 transactions the sweep is 49 to 50 live rows in 13 ms per block, where the old code would have spent about 33 s. The live set tracked the immature coinstakes exactly as stakes landed and matured.

## Verification

- `/proc` sampling of the GUI thread at 1 Hz across blocks, before and after, aligned with `UpdateTip`; figures above.
- `-debug=bench` lines per wallet per block, before and after.
- In the running client: labels shown in the history view and the overview list, search matching on address, label and txid, a new stake appearing as a row and losing its brackets at maturity, immature rows leaving the live set at 31 confirmations.
- `test_reddcoin-qt` aborts in `WalletTests` on this line (RED-122), so unit coverage lands with the `develop` port.

Behaviour to note: a label edit now repaints in the history view on the next repaint rather than on the next block.

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -14,6 +14,7 @@
 
 #include <QFont>
 #include <QDebug>
+#include <QHash>
 
 const QString AddressTableModel::Send = "S";
 const QString AddressTableModel::Receive = "R";
@@ -70,6 +71,12 @@ class AddressTablePriv
 {
 public:
     QList<AddressTableEntry> cachedAddressTable;
+    /** Label of every non-change address book entry, keyed by the address
+     *  as EncodeDestination spells it. Unlike cachedAddressTable this is
+     *  never narrowed by pk_hash_only, so labelForAddress() can answer from
+     *  it for any address without taking the wallet lock. Kept current by
+     *  the same notifications that maintain the table. */
+    QHash<QString, QString> labels;
     AddressTableModel *parent;
 
     explicit AddressTablePriv(AddressTableModel *_parent):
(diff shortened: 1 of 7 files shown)


reddink:
Port to develop, with the unit tests that this line cannot run: #559.

PR #560, opened 2026-09-11
staker: publish the stake weight from the search pass

reddink:
With the transaction-table refresh fixed in #558, the stake-weight recomputation was all that remained of the per-block GUI stall, and it grows with the wallet's transaction count. Tracked as [RED-131](https://reddink.youtrack.cloud/issue/RED-131), part of the [RED-130](https://reddink.youtrack.cloud/issue/RED-130) epic.

## The problem

`WalletModel::GetStakeWeight` recomputed the wallet's stake weight from scratch on every call: a hard `cs_wallet` lock, a full balance scan, a full available-coins scan, a complete coin selection for the whole balance, and then a txindex lookup and block-file read per UTXO. The status bar called it on every block and on every staking notification; the node window's staking tab called it on every block-tip signal, headers included, whether or not the window was open. A block this wallet found cost four calls on the GUI thread. `getstakinginfo` ran the same scan under `cs_wallet` on every poll.

Measured on `--enable-debug` builds:

| Wallet | Transactions | One call | GUI thread per found block |
|---|---|---|---|
| Main_funds, mainnet | 176k | 0.8 s | 1.5 to 3 s |
| staketest, testnet | 1.18M | 5.8 s | 19 to 26 s at 100% |

## The change

The staking thread already examines every stakeable coin on every pass and holds, per coin, the value and the timestamp the weight formula needs. `CreateCoinStake` now sums the weight as it goes, before its age gate so the sum matches the on-demand formula, and returns it in a `StakeWeightSummary`. `CreateNewBlock` publishes it on the wallet beside the search interval it already records, as two atomics and a known flag. `WalletModel::GetStakeWeight`, the node window and `getstakinginfo` read the published values, which takes no lock. The staking thread resets them on exit.

Two states are distinguished so the callers can act on them. A pass that finds nothing stakeable publishes a known zero, and the status bar reports "no mature coins". Before any pass has completed, and after the thread stops, the value is unknown, and the status bar reports "waiting for staking to start"; `getstakinginfo` computes on demand in that case, so it stays meaningful for a wallet that is not staking.

The search still stops at the first kernel, but the sum continues over the rest of the set, and every pass that runs to the end publishes. A wallet large enough to find a kernel on every pass, which the testnet wallet is, would otherwise never publish; the first cut of this change had that bug and testnet found it.

The GUI's stale-tip check used to try `cs_main` and give up if busy. The staking thread holds `cs_main` for a good part of every pass, so a status refresh landing during one reported no weight and showed "waiting" until the next block. It now reads the tip time from the client model's cache, which the tip notifications already keep.

The coin-set overload of the weight function, `Node::getStakeWeight`, `CWallet::GetStakeWeightSet` and `Node::tryGetSyncInfo` had no other callers and are removed.

## Result

Testnet, staketest, 1,182,446 transactions, 461 UTXOs, staking, five minutes with five blocks all found by this wallet:

| | Before | After |
|---|---|---|
| `getstakinginfo` while staking | 6.46 s under `cs_wallet` | 0.00 s |
| GUI thread after a block this wallet found | 19 to 26 s at 100% | 1 s |
| GUI thread over 300 s | | 294 s idle, 5 s computing, 0 s blocked on a lock |

The remaining second per block on this wallet is the balance poll's scan of 1.18M transactions (RED-135) and the minting-table update (RED-136).

## Tests

`wallet_tests/published_stake_weight` pins the contract the callers depend on: unknown before any pass, known after one including a known zero, unknown again after the thread stops, with the outputs cleared in the unknown case.

The published-equals-computed check needs a harness that can stake, which this line's cannot; it lands with the `develop` port.

Verified in the running testnet client: `getstakinginfo` answers instantly with the published weights, and the status bar and the node window's staking tab show the same weights as the RPC.

### DIFF
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -26,7 +26,6 @@
 class BanMan;
 class CCoinControl;
 class CFeeRate;
-class CInputCoin;
 class CNodeStats;
 class Coin;
 class RPCTimerInterface;
@@ -151,9 +150,6 @@ public:
 
     //! Get PoSVKernelPS.
     virtual uint64_t getPoSVKernelPS() = 0;
-
-    //! Get the total and average weights from the wallet for staking.
-    virtual bool getStakeWeight(std::set<CInputCoin>& setCoins, uint64_t& nAverageWeight, uint64_t & nTotalWeight) = 0;
 
     //! Set staking active.
     virtual void setNodeStakingActive(bool active) = 0;
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -193,9 +193,6 @@ public:
     //! Get node synchronization information.
     virtual void getSyncInfo(int& numBlocks, bool& isSyncing) = 0;
 
-    //! Try to get node synchronization information.
-    virtual bool tryGetSyncInfo(int& numBlocks, bool& isSyncing) = 0;
-
     //! Get wallet client.
     virtual WalletClient& walletClient() = 0;
 
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -25,7 +25,6 @@
 
 class CCoinControl;
 class CFeeRate;
-class CInputCoin;
 class CKey;
 class CWallet;
 enum class FeeReason;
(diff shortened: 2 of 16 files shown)


reddink:
Port to develop, with the seam test that this line cannot run: #561.

PR #562, opened 2026-09-11
wallet: stop asking the chain whether a confirmed transaction is final

reddink:
Every wallet balance and coin scan asked the chain, once per transaction, whether that transaction was final. Each call took `cs_main` and recomputed the tip's median time past, and the answer never changed the result. Tracked as [RED-135](https://reddink.youtrack.cloud/issue/RED-135), part of the [RED-130](https://reddink.youtrack.cloud/issue/RED-130) epic.

## The problem

`IsTrusted` called `chain().checkFinalTx` before it looked at anything else, and `AvailableCoins` called it on its own account and then again through `IsTrusted`. `GetBalance` runs `IsTrusted` per transaction, so the GUI balance poll, `getbalances`, and every staking pass paid one to three `cs_main` acquisitions per wallet transaction, each also computing an eleven-block median. On a wallet with hundreds of thousands of transactions that was most of the cost of the scan, and every acquisition waited behind whatever else held `cs_main`, the staking thread included, which holds it for a good part of every pass.

Measured on `--enable-debug` builds before this change: `getbalances` 0.17 s at 176k transactions and 0.95 s at 1.18M; `listunspent` 0.87 s at 176k. The balance poll ran that scan once per wallet per block on the GUI thread.

## The change

Finality only grows with the chain. A transaction at depth one or more passed the check when it was mined and passes at every later tip; an unconfirmed transaction the mempool holds passed the same check on acceptance; an unconfirmed transaction outside the mempool is refused by both scans on that ground alone. So the check is gone from `IsTrusted` and `AvailableCoins`, with the depth and mempool tests that were already there carrying the result. The two received-by RPC tallies test depth first and ask the chain only below depth one.

The one place finality carries information is the GUI's "open until" status for a locktime transaction the wallet holds but the mempool does not. `MakeWalletTxStatus` asks the chain in that case only.

Later Bitcoin Core releases removed the same calls from the same two functions on the same reasoning.

## Result

Testnet, `staketest`, 1,182,779 transactions, 459 UTXOs, `--enable-debug` build:

| | Before | After |
|---|---|---|
| `getbalances`, warm | 0.95 s | 0.60 s |
| `getstakinginfo`, on-demand path for a wallet that is not staking | 5.76 s | 4.98 s |

What remains of the balance scan is walking 1.18M map entries and their cached credits. `listunspent` (4.9 s on this wallet) and the on-demand weight are dominated by the per-output `IsSpent` lookup and script solve, which is O(outputs) work rather than a lock and does not sit on the per-block GUI path.

GUI thread over three minutes with staking on and three blocks found by this wallet: 4 s computing, 13 s blocked on a lock, 174 s idle. The computing is two or three balance scans per found block, since each transaction notification re-raises the poll's recheck flag (RED-137). The blocked time sits exactly on the staker's passes, which the GUI's remaining hard wallet locks wait out, and is now the largest stall left on a large wallet (RED-133, RED-132).

## Tests

`wallet_tests/trusted_and_available_without_finality_check` builds three transactions by hand on the fixture chain and pins every case the removed check could have told apart: a confirmed transaction is trusted and final with its unspent output available; an unconfirmed spend outside the mempool is neither trusted nor available whether or not it is final, and becomes both once the wallet is told the mempool holds it; a locktime spend a hundred blocks ahead, with a non-final sequence so the locktime binds, is never trusted or available and still reports not final.

Passes. Verified in the running testnet client: balances, `listunspent` and the history view unchanged; the client staked through the measurement window.

### DIFF
--- a/src/wallet/interfaces.cpp
+++ b/src/wallet/interfaces.cpp
@@ -86,7 +86,11 @@ WalletTxStatus MakeWalletTxStatus(const CWallet& wallet, const CWalletTx& wtx)
     result.depth_in_main_chain = wtx.GetDepthInMainChain();
     result.time_received = wtx.nTimeReceived;
     result.lock_time = wtx.tx->nLockTime;
-    result.is_final = wallet.chain().checkFinalTx(*wtx.tx);
+    // Finality is settled for a confirmed transaction and for one the mempool
+    // accepted; only an unconfirmed transaction the mempool does not hold, a
+    // locktime transaction the wallet is keeping until it is final, needs the
+    // chain asked, and that is the one case the GUI shows as "open until".
+    result.is_final = result.depth_in_main_chain >= 1 || wtx.InMempool() || wallet.chain().checkFinalTx(*wtx.tx);
     result.is_trusted = wtx.IsTrusted();
     result.is_abandoned = wtx.isAbandoned();
     result.is_coinbase = wtx.IsCoinBase();
--- a/src/wallet/receive.cpp
+++ b/src/wallet/receive.cpp
@@ -280,8 +280,15 @@ void CWalletTx::GetAmounts(std::list<COutputEntry>& listReceived,
 bool CWallet::IsTrusted(const CWalletTx& wtx, std::set<uint256>& trusted_parents) const
 {
     AssertLockHeld(cs_wallet);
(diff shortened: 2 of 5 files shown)


reddink:
Port to develop: #563.

PR #564, opened 2026-09-12
staker: search for a kernel without holding the wallet lock

reddink:
Tracked as [RED-133](https://reddink.youtrack.cloud/issue/RED-133), part of the [RED-130](https://reddink.youtrack.cloud/issue/RED-130) epic. Follows #562.

## The problem

The staking thread took `cs_wallet` around `CreateNewBlock` for every search pass, and `CreateCoinStake` inside it took `cs_main` as well. The pass itself is the long part: on a large wallet it walks every stakeable coin, reads each one's source block from disk through the transaction index, and runs up to sixty kernel hashes per coin. With RED-134, RED-131 and RED-135 in, that pass had become the largest stall left on the GUI thread: over three minutes on the 1.18M transaction testnet wallet the GUI spent 4 s computing and 13 s blocked on a lock, and every one of the blocked seconds sat on a staker pass.

None of that work needs the wallet lock. The coin set only changes when the wallet processes a block, the disk reads go through the index, and the kernel check needs `cs_main` only for the moment it reads the chain.

## The change

`CreateCoinStake` is split into three steps in `pos/stake.cpp`:

- `CollectStakeCandidates` takes `cs_wallet` and records the wallet's last processed block, its trusted balance, the reserve, and the selected coins. This is the scan that used to run on every pass; the staking thread now runs it once, and again only when the wallet's last block moves or a kernel fails its re-check.
- `SearchStakeKernel` walks the candidates with no wallet lock, reads each coin's source through the index, holds `cs_main` around each kernel check alone, and reports the stake weight of the coins it examined the way the pass already did for RED-131. A keyhash kernel needs one key lookup, taken under a short `cs_wallet` on its own, never nested under `cs_main`.
- `BuildCoinStake` requires both locks and first re-checks the kernel against the tip the block will sit on: unspent in the wallet, unspent on the chain, still mature, and still meeting the target. Then it builds the transaction exactly as before. Combine inputs are checked for spends too, since the candidates may predate the build.

`CreateCoinStake` runs the three in turn under `LOCK2(cs_main, cs_wallet)`, so the generate RPCs behave as they did. `CreateNewBlock` takes an optional candidate set and kernel; given one, it builds around the kernel instead of searching.

`PoSMiner` collects, searches lock-free each pass, publishes the search interval and the weight from the search, and takes `cs_wallet` for the block only once a kernel has been found. A pass that finds nothing never takes either lock. A refused kernel marks the candidates for re-collection so a coin the wallet spent between blocks costs one pass, not a block interval. The search clock is per thread rather than the static shared with the RPCs, and the sleep between passes is sized from the collected candidates instead of a separate coin scan at startup.

## Result

Testnet, `staketest`, 1,182,900 transactions, 459 UTXOs, staking on, `--enable-debug` build. GUI thread sampled at 1 Hz from `/proc`:

| | Before (#562) | After, 300 s window |
|---|---|---|
| Blocks found by this wallet in the window | 3 in 180 s | 4 in 300 s |
| GUI seconds blocked on a lock | 13 | 0 |
| GUI CPU per block found | 0.7 to 1.9 s | 0.6 to 0.7 s |
| Staker thread duty cycle | 11.6 % | 7.4 % |

Four windows were sampled and none held a single lock-wait sample, including the five seconds after each found block during which the thread re-collects its candidates under `cs_wallet`: the GUI's own per-block work no longer takes a hard wallet lock, so nothing waits on it. What remains per found block is the balance rescan that follows a new wallet transaction, which is RED-137.

Staking itself: 59 blocks found and 59 accepted as tip in the first 66 minutes, one every 67 to 68 seconds, zero kernel refusals, signing failures or rejected blocks. `getstakinginfo` reports the published weight and a moving search interval.

## Tests

`wallet_tests/stake_candidates_searched_without_wallet_lock`. Regtest cannot stake in this harness, so it pins the contract between the three steps instead: a collection records the wallet's last block, balance and coins in search order, and reports an uncoverable reserve as done with nothing to stake; a search over an empty collection reports a known zero weight; and the build refuses a kernel the wallet has spent since the collection, one the chain does not hold, and one the chain no longer counts mature, each identified by its log line, directly and through the block assembler, which reports it as a cancelled pass.

The other five `wallet_tests` failures on this line predate the branch and fail identically on a binary built at the #562 merge; they are the Bitcoin fixture numbers tracked in RED-125.

## Notes

- Lock order on this line is `cs_wallet` before `cs_main`, from the staking thread, the generate RPC and the existing wallet tests. The new code never takes `cs_wallet` under `cs_main`, and the measured client is a `DEBUG_LOCKORDER` build.
- A coin the wallet spends between blocks is only noticed at the build, which then re-collects. Locked coins still make the whole-balance coin selection fail; that is pre-existing.
- The develop port is not a cherry-pick: develop has no `pos/stake.cpp` and the block assembler reaches the wallet through `interfaces::StakingWallet`, so the split lands in `wallet/staking.cpp` there.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -140,7 +140,7 @@ void BlockAssembler::resetBlock()
     nFees = 0;
 }
 
-std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& scriptPubKeyIn, CWallet* pwallet, bool* pfPoSCancel)
+std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& scriptPubKeyIn, CWallet* pwallet, bool* pfPoSCancel, const StakeCandidates* candidates, const StakeKernel* kernel)
 {
     int64_t nTimeStart = GetTimeMicros();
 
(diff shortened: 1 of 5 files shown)


reddink:
develop port: #565. Not a cherry-pick, see its description for what differs on that line.

